PR #8, opened 2020-09-13
Feature/typings to september 12

santee:


skalkin:
Thank you Alexander! We are planning to move to TypeScript soon, so hopefully, this is the last time you are doing this :) We appreciate it a lot!

PR #20, opened 2021-02-17
get_aligned_coords now also works on scaffolds with R-groups and returns substructure match

ptosco:
* bumped RDKit_minimal version number to 2021.03_05
* simplified _molIsInMolBlock function to detect if the passed molString is a molblock and renamed to _isMolBlock
* now generate_aligned-coords will return the matched substructure JSON if successful and empty string if not, so no need to to call get_substruct_match beforehand
* get_new_coords is used to generate new coordinates rather than roundtripping through SMILES. This is more efficient and ensures that CoordGen is used for coordinate generation


packages/Chem/src/rdkit_cell_renderer.js
```diff
@@ -50,26 +41,14 @@ class RDKitCellRenderer extends DG.GridCellRenderer {
...
           let rdkitScaffoldMol = this._fetchMol(scaffoldMolString, "", molRegenerateCoords, false).mol;
-          if (this._molIsInMolBlock(scaffoldMolString, rdkitScaffoldMol)) {
-            substructJson = mol.get_substruct_match(rdkitScaffoldMol);
-            if (substructJson !== '{}') {
-              mol.generate_aligned_coords(rdkitScaffoldMol, true);
-            }
-          }
+          substructJson = mol.generate_aligned_coords(rdkitScaffoldMol, true, true);
+        } else if (molRegenerateCoords) {
+          mol = rdKitModule.get_mol(mol.get_new_coords(true));
```
dskatov, line 48: @ptosco should we `mol.delete()` before this to not let memory leak?

ptosco: @dskatov 
Good catch (https://emscripten.org/docs/porting/connecting_cpp_and_javascript/embind.html#memory-management)
I have just submitted another PR to fix this:
https://github.com/datagrok-ai/public/pull/21
Could you please merge it for me? Thanks a lot.

PR #31, opened 2021-04-10
Fixes CoordGen regression

ptosco:
Fixes the `CoordGen` regression that was causing 10x slower performance compared to the past.
Before this fix:
```
%%time
rdDepictor.SetPreferCoordGen(True)
for m in chembl_mols_2000:
    rdDepictor.Compute2DCoords(m)
CPU times: user 18min 10s, sys: 868 ms, total: 18min 11s
Wall time: 18min 10s
```
After:
```
%%time
rdDepictor.SetPreferCoordGen(True)
for m in chembl_mols_2000:
    rdDepictor.Compute2DCoords(m)
CPU times: user 2min 5s, sys: 53 ms, total: 2min 5s
Wall time: 2min 5s
```


skalkin:
Awesome, thanks Paolo!

PR #39, opened 2021-07-13
Normalize and prettify 2D layouts

ptosco:
- 2D layouts are now normalized wrt their orientation
- F, Cl, and S elements are more visible on a green selection background
- maximum size of molecules is clipped
- font/molecule size ratio is adjusted
- the latest `coordgenlibs` master was picked to fix some 2D layout distortions that we occasionally found

@dskatov Before you jump on your chair when you see this PR: the whole idea behind this is to have molecules nicely oriented along their principal axes and biased in their rotation such that the majority of bonds have a 30 degree angle with the X axis, as they look much prettier. CoordGen already does this.
The orientation is triggered for depictions from SMILES (tested on a local ChEMBL file on plain DG) at the same time as the depiction is created.
For our use case where we load 2D molblocks into the `Structure` column, upon initialization of the SubstructureLibrary I assign to each of the workers a slice of molecules to orient. Once all workers are done, the original molblocks are overwritten.
As a bonus, molblocks whose coordinates differ of a few decimals are now consolidated, such that categorization of molecules based on molblock identity is not fooled by irrelevant differences.
I know that implementing the reorientation as part of the SubstructureLibrary initialization is a hack, but as I do not have access to the DG sources that was the only way for me to test that it was working - you may want to move it elsewhere to make it prettier.


ptosco:
@dskatov I have just pushed an updated MinimalLib where I have implemented `get_morgan_fp_as_uint8array()`, which takes the same parameters as `get_morgan_fp` but returns a JS `Uint8Array` instead of a `string`.

dskatov (commented):
Reviewed and accepted

PR #88, opened 2021-08-26
Custom ML Functions

GitDenisGostev:


alex-aprm (approved):

alex-aprm:
@konstantin-doncov what do you think?

packages/CustomML/src/package.js
```diff
@@ -0,0 +1,11 @@
...
+import * as grok from 'datagrok-api/grok';
+import * as ui from 'datagrok-api/ui';
+import * as DG from 'datagrok-api/dg';
+
+export let _package = new DG.Package();
+
+//name: test
+export function test() {
+  grok.shell.info(_package.webRoot);
+}
```
konstantin-doncov, line 11: Better to remove this test function

konstantin-doncov (changes requested):
Please remove the test function

packages/CustomML/src/package.js
Denomet: Removed test function from package.js

Issue #141, opened 2021-10-14
Chem: porting "Similarity Analysis" to TS


StLeonidas:
Port Similarity Analysis from core to Chem package in TypeScript

PR #177, opened 2021-11-03
Chem: Similarity Analysis, BitArray

nuradilK:
https://github.com/datagrok-ai/public/issues/141


packages/Chem/package.js
```diff
@@ -141,6 +141,42 @@ class ChemPackage extends DG.Package {
     }
   }
 
+  //name: chemSimilaritySpeImpl
+  //input: column fingerprints [fingerprints]
+  //input: int dimension
+  //output: dataframe result
+  async chemSimilaritySpeImpl(fingerprints, dimension) {
+    var core = _fingerprintSimilarity;
+    var coordinates = stochasticProximityEmbedding(fingerprints.length, (x, y) => 1.0 - core(fingerprints.get(x), fingerprints.get(y)),
```
skalkin: Where are we passing the number of dimensions?

packages/Chem/package.js
```diff
@@ -141,6 +141,42 @@ class ChemPackage extends DG.Package {
...
+  }
+
+  //name: chemSimilarityAnalysis
+  //input: dataframe table
+  //input: column molColumn {semType: Molecule}
+  //output: graphics
+  async chemSimilarityAnalysis(table, molColumn) {
+    var fpColumn = await this.getMorganFingerprints(molColumn);
+    if (fpColumn.stats.missingValueCount > 0) {
+      throw "Molecule column has a null entry";
```
skalkin: That shouldn't be an error - these rows should simply be excluded

packages/Chem/package.js
nuradilK: This is an old commit, pls check the recent ones. I already removed this function.

Issue #93, opened 2021-08-28
Peptides: peptide space


skalkin:
Visualize a collection of peptides in 2-dimensional space, using [t-SNE](https://en.wikipedia.org/wiki/T-distributed_stochastic_neighbor_embedding) with the [edit distance](https://en.wikipedia.org/wiki/Edit_distance) function. Implement Peptide space feature similar to Chemical space feature in Chem package:

- [x] Implement an algorithm to calculate similarities for a pair of sequences (semtype - alignedSequence)
- [x] Use SPE algorithm to get 2d map of peptide sequences
- [x] Implement a new feature in the property panel of each column with alignedSequence values 
- [x] Implement embedding as a web worker to not slow down the main thread.
- [x] Benchmark UMAP performance on dataset of 40k sequences.

StLeonidas:
Implement Peptide space feature similar to Chemical space feature in Chem package:
- [ ] Implement an algorithm to calculate similarities for a pair of sequences (semtype - alignedSequence)
- [ ] Use SPE algorithm to get 2d map of peptide sequences
- [ ] Implement a new feature in the property panel of each column with alignedSequence values

nikolay-alemasov:
Do we really need to consider only alignedSequence? As I can remember string metrics can work on strings of different length as well.

nikolay-alemasov:
Only this left:

1. Implement a new feature in the property panel of each column with alignedSequence values

nikolay-alemasov:
I would also add the following task:

- [ ] Implement embedding as a web worker to not slow down the main thread.

nikolay-alemasov:
In fact, the Peptide Space could be considered as ready with some minor bugs.

nikolay-alemasov:
- [ ] Benchmark UMAP performance on dataset of 40k sequences.

nikolay-alemasov:
Suggest to close the issue.

PR #206, opened 2021-11-18
Utils, Peptides: Peptide space now use fixed utils

nikolay-alemasov:
### Implements

#93 

### Changed

1. Utils subroutines now moved back to in-house implementation instead of matjs of vectorous.
2. Peptide space changed to use Utils modified.
3. JSDoc added.


packages/Peptides/src/package.ts
```diff
@@ -344,73 +352,58 @@ export function peptideSimilaritySpace(
   alignedSequencesColumn: DG.Column,
   method: string,
   measure: string,
-  cyclesCount: number) {
+  cyclesCount: number)
+// eslint-disable-next-line brace-style
```
vdyma, line 356: Do not use `eslint-disable-next-line`, fix the error instead

packages/Peptides/src/package.ts
```diff
@@ -344,73 +352,58 @@ export function peptideSimilaritySpace(
...
-export async function testPeptideSimilaritySpace() {
-  const v1 = 'ABCDEF';
-  const v2 = 'BCDEFGH';
-  const d = new Measurer('Jaro-Winkler').getMeasure()(v1, v2);
-  console.log('testPeptideSimilaritySpace:');
-  console.log(v1);
-  console.log(v2);
-  console.log(d);
-  return;
+// eslint-disable-next-line no-unused-vars
```
vdyma, line 404: Do not use eslint-disable-next-line, fix the error instead

vdyma (approved):
Looks good except for a couple of places where the eslint is set to ignore errors. Thanks for your work, @nikolay-alemasov!

PR #219, opened 2021-11-29
Documentation: Correlation plot can also show Spearman's correlation.

nikolay-alemasov:
Added link to the Spearman's correlation coefficient which can also be selected by an user.


nikolay-alemasov:
Also, please, consider #220

PR #231, opened 2021-12-15
Simplify RDKit MinimalLib updates

ptosco:
This PR implements two changes:

- It removes the need to patch the `RDKit_minimal.js` loader whenever we update the `wasm` library version by passing the `locateFile` to the `initRDKitModule` function. This allows to keep the `RDKit_minimal.js` loader as generated by `emscripten` - no need to patch it anymore
- The `RDKit_minimal` library version is stored in a single place for ease of updating while previously multiple files had to be modified at each update.
To keep things simple I have removed the versioning information from the JS loader library (which indeed is not changing when we update the `wasm` library, particularly as we do not need to patch it anymore thanks to this PR)


dskatov:
Thank you for the PR @ptosco!

PR #250, opened 2022-01-07
Chem: Switch Chemical Space to use UMAP rather than SPE

nikolay-alemasov:
Just switched Chemical Space to UMAP embedding from SPE. Needs further investigation to find out why SPE fails.


skalkin (approved):
Looks good to me if it's working

PR #282, opened 2022-01-20
ML: Fix SPE function and add new similarity metrics

nuradilK:


libraries/ml/src/spe.ts
```diff
@@ -176,3 +176,89 @@ export class PSPEBase extends SPEBase {
...
+    // Initialize the D-dimensional coordinates of the N points.
+    const coordinates = fillRandomMatrix(nItems, OriginalSPE.dimension, areaWidth);
+
+    this.initDistance(vectors);
+
+    if (this.maxDistanceSteps == null) {
+      this.maxDistanceSteps = nItems * Math.floor((nItems - 1) / 2);
+    }
+    if (this.maxDistance == null) {
+      this.maxDistance = -1e37;
```
nikolay-alemasov, line 219: 1. What is the constant for? 
2. And why is it equals that number? 
3. Why is it declared within `this` scope?

libraries/ml/src/spe.ts
```diff
@@ -176,3 +176,89 @@ export class PSPEBase extends SPEBase {
...
+        const rowi = coordinates[i]; const rowj = coordinates[j];
+
+        // ... retrieve (or evaluate) their proximity in the input space, rij and ...
+        const r = this.calcDistance(vectors, i, j);
+        // ... compute their Euclidean distance on the D-dimensional map, dij.
+        const d = calculateEuclideanDistance(rowi, rowj);
+
+        // If rij <= rc, or if rij > rc and dij < rij ...
+        if ((r <= radius) || (d < r)) {
+          const multiplier = lambda*0.5*(r-d)/(d+this.epsilon);
```
nikolay-alemasov, line 249: We should revise spaces between operations. It should be like:
`          const multiplier = lambda * 0.5 * (r - d) / (d + this.epsilon);
`

libraries/ml/src/spe.ts
nuradilK: I believe `maxDistanceSteps` parameter is needed to find the most optimal `maxDistance` parameter. I am not the author of this algorithm. I just referred to the algorithm from the core. I can send you the original code, if you want. Also, `maxDistanceSteps` in `this` scope, because it could be set in the constructor.

libraries/ml/src/string-measure.ts
```diff
@@ -2,12 +2,36 @@ import * as fl from 'fastest-levenshtein';
 import {jaroWinkler} from 'jaro-winkler-typescript';
 
 import {DistanceMetric} from '@datagrok-libraries/utils/src/type-declarations';
+import * as metric from '@datagrok-libraries/utils/src/similarity-metrics';
```
nikolay-alemasov, line 5: It should not be included here. There should be more general class including both string and chemical measures in this ML library not to mix different entities.

libraries/utils/src/similarity-metrics.ts
```diff
@@ -0,0 +1,83 @@
...
+export function rogotGoldbergSimilarity(x: BitArray, y: BitArray) {
+  let common = x.andWithCountBits(y, true);
+  let total = x.countBits(true) + y.countBits(true);
+  let len = x.length;
+  let diff = len - total + common;
+  if ((common == len) || (diff == len))
+    return 1.0;
+  else
+    return common / total + diff / (2 * len - total);
+}
```
nikolay-alemasov, line 83: No new line

nikolay-alemasov (changes requested):
Please consider my comments before I can approve.

PR #336, opened 2022-02-21
API: changed InputBase input() return type

AufarZakiev:
InputBase's `input()` method returns the actual HTML element under the hood of InputBase. It is always HTMLInputElement with particular interface


AufarZakiev:
Ok, I will delete PR soon

PR #554, opened 2022-05-04
Update DG to RDKit_minimal_2022.03_1

ptosco:
- update DG to RDKit_minimal_2022.03_1
- add support for chemical reaction drawing
- add support for additional FPs


dskatov:
@ptosco thank you for the update!

skalkin:
Awesome, thank you Paolo!

dskatov:
@ptosco, we've figured out that the alignment through:
```
mol.normalize_2d_molblock();
mol.straighten_2d_layout();
```
worked in `RDKit_minimal_2021.03_18.wasm`, but no longer works in `RDKit_minimal_2022.03_1.wasm` with:
```
mol!.normalize_depiction();
mol!.straighten_depiction();
```
What could be the reason, could you please check?

ptosco:
@dskatov This should be fixed by #562.

PR #562, opened 2022-05-05
Integrates #554

ptosco:
Integrates #554.
Make sure the molecule has coordinates before calling `normalize_depiction()` and `straighten_depiction()`


dskatov:
@ptosco thanks, this fixes it.

PR #683, opened 2022-06-06
Dkryvchuk/chem tests

dionissqq:


skalkin (approved):
Looks good to me. Please provide more meaningful comments though, they should reflect the package and also start with an upper case :) This would be better:

`Chem: Tests: more tests on substructure search`

PR #684, opened 2022-06-06
Dkryvchuk/642 mol col autodetection

dionissqq:


packages/Chem/detectors.js
```diff
@@ -1,49 +1,30 @@
...
     if (col.type !== DG.TYPE.STRING)
       return null;
-
-    if (col.toList().every((el) => el && el.length > 0 &&
-      ChemPackageDetectors.sValidFirstSmilesChar.includes(el[0]) &&
-      el.split('').every((sChar) => ChemPackageDetectors.sValidSmilesChar[sChar.charCodeAt(0)]) == 1)) {
+    console.log('hmmm');
+    if (col.toList().every((el) => el && el.length > 1 &&
+      ChemPackageDetectors.validFstChar[el[0]] &&
+      el.split('').every((sChar) => ChemPackageDetectors.validChar[sChar]))) {
```
skalkin: Why would you want to create an array of one-char strings? This is super-wasteful

packages/Chem/detectors.js
```diff
@@ -1,49 +1,30 @@
...
-  static sValidFirstSmilesChar = ['(', 'B', 'C', 'D', 'F', 'I', 'N', 'O', 'P', 'S', '[', 'c', 'n', 'o'];
+  static sValidFirstSmilesChar = Uint32Array.from(
+    Buffer.from('AAAAAACAAAA6Q5AQEAMQAAAAAAAAAAAAAAAAAAAAAAA', 'base64'));
+
+  static validChar(char) {
+    const pos = char.charCodeAt(0);
+    return (this.sValidSmilesChar[Math.floor(pos / 0x20)] & (1 << (pos & 0x1f))) != 0;
+  }
+
+  static validFstChar(char) {
```
skalkin, line 11: No need to shorten "first" to "fst"

packages/Chem/detectors.js
```diff
@@ -21,18 +19,19 @@ class ChemPackageDetectors extends DG.Package {
   detectRDSmiles(col) {
     if (col.type !== DG.TYPE.STRING)
       return null;
-    console.log('hmmm');
+    // return DG.SEMTYPE.MOLECULE;
     if (col.toList().every((el) => el && el.length > 1 &&
```
skalkin, line 23: You have just created an extra array for no good reason, keep in mind that this code should be as performant as possible

packages/Chem/src/package.ts
```diff
@@ -151,6 +151,27 @@ export async function getMorganFingerprints(molColumn: DG.Column): Promise<DG.Co
...
+//output: double perc
+export async function checkSmilesValidity(col: DG.Column): Promise<number> {
+  if (col.length == 0)
+    return 0;
+  let valid = 0;
+  let invalid = 0;
+  const inc = col.length < 20 ? 1 : Math.floor(col.length / 20);
+  for (let i = 0; i < col.length; i += inc) {
+    try {
+      getRdKitModule().get_mol(col.get(i));
```
skalkin, line 165: `get_mol` returns a C++ molecule, which should be disposed of

packages/Chem/src/package.ts
```diff
@@ -151,6 +151,27 @@ export async function getMorganFingerprints(molColumn: DG.Column): Promise<DG.Co
   }
 }
 
+//name: checkSmilesValidity
+//input: column col
+//output: double perc
+export async function checkSmilesValidity(col: DG.Column): Promise<number> {
```
skalkin, line 157: No reason to check the whole column, this will take too much time, we have agreed on just passing 10 random values

skalkin (changes requested):
Please address the comments I've made, and then re-submit the PR.

packages/Chem/src/package.ts
```diff
@@ -151,6 +151,25 @@ export async function getMorganFingerprints(molColumn: DG.Column): Promise<DG.Co
   }
 }
 
+//name: checkSmilesValidity
+//input: list molStrings
+//output: double perc
+export async function checkSmilesValidity(molStrings: string[]): Promise<number> {
```
skalkin, line 157: We agreed to stop iterating once it's clear you can't reach the threshold

packages/Chem/detectors.js
```diff
@@ -18,21 +21,47 @@ class ChemPackageDetectors extends DG.Package {
...
     }
 
-    if (DG.Detector.sampleCategories(col, (s) => ChemPackageDetectors.likelySmiles(s)))
-      return DG.SEMTYPE.MOLECULE;
+    let molStrings = [];
+    if (col.length <= ChemPackageDetectors.numberOfMolsToCheck)
+      molStrings = col.toList().filter(el => el);
+    else {
+      const inc = Math.floor(col.length / ChemPackageDetectors.numberOfMolsToCheck);
+      for (let i = 0; i < col.length; i += inc) {
```
skalkin, line 41: You are iterating to col.length and not to inc

packages/Chem/detectors.js
```diff
@@ -18,21 +21,47 @@ class ChemPackageDetectors extends DG.Package {
...
+          if (!ChemPackageDetectors.validChar(el[j]))
+            return null;
+      }
     }
 
-    if (DG.Detector.sampleCategories(col, (s) => ChemPackageDetectors.likelySmiles(s)))
-      return DG.SEMTYPE.MOLECULE;
+    let molStrings = [];
+    if (col.length <= ChemPackageDetectors.numberOfMolsToCheck)
+      molStrings = col.toList().filter(el => el);
```
skalkin, line 38: *Never* use toList in semType detectors without a compelling reason. Cloned arrays are super wasteful!

packages/Chem/detectors.js
```diff
@@ -18,21 +21,47 @@ class ChemPackageDetectors extends DG.Package {
...
+    else {
+      const inc = Math.floor(col.length / ChemPackageDetectors.numberOfMolsToCheck);
+      for (let i = 0; i < col.length; i += inc) {
+        let el = col.get(i);
+        if (el)
```
skalkin, line 43: You were already doing this check on line 37

packages/Chem/detectors.js
skalkin: Oh, I see, you were doing the check in the other if branch. Try to simplify the code

packages/Chem/detectors.js
```diff
@@ -18,21 +21,47 @@ class ChemPackageDetectors extends DG.Package {
...
-      // smiles or molblock?
-      let str = col.length > 0 ? col.get(0) : null;
-      col.tags[DG.TAGS.UNITS] = (str !== null && str.includes('M  END')) ? DG.UNITS.Molecule.MOLBLOCK : DG.UNITS.Molecule.SMILES;
-
-      return col.semType;
+    // return DG.SEMTYPE.MOLECULE;
+    for (let i = 0; i < col.length; ++i) {
+      const el = col.get(i);
+      if (el && el.length > 0) {
+        if (!ChemPackageDetectors.validFirstChar(el[0]))
```
skalkin, line 28: I have a feeling it would be a lot more efficient to use charCodeAt() instead of creating a one-character string

skalkin (changes requested):

packages/Chem/detectors.js
dionissqq: it uses charCodeAt() inside validFirstChar() function

skalkin (changes requested):
Looks good, but let's address the latest comments before merging it

Issue #1125, opened 2022-10-17
JS API: add methods for getting value color into ColumnColorHelper


onuf:
```ts
export class ColumnColorHelper {
  getColor(i: number): number {}
  getColors(): UInt32Array {}
}
```

onuf:
@skalkin it looks like we need one more parameter, unless we want to assume something about the grid (e.g., find the view for the dataframe via grok.shell and use view.grid) or move the methods to GridColumn.
```ts
  getColor(grid: Grid, i: number): number {
    return grid.cell(this.column.name, i).color;
  }

  getColors(grid: Grid): Uint32Array {
    const colors = new Uint32Array(this.column.length);
    for (let i = 0; i < this.column.length; i++)
      colors[i] = this.getColor(grid, i);
    return colors;
  }

// grid will most likely come from view.grid (viewer.view.grid and so on)
t.col('age').meta.colors.getColors(grid);
t.col('age').meta.colors.getColor(grid, 5);
```

onuf:
```ts
// usage
t.col('name').meta.colors.getColors();
t.col('name').meta.colors.getColor(5);
```

PR #1137, opened 2022-10-20
#1125 add methods for getting value color into ColumnColorHelper

onuf:


onuf:
Initial signatures:
```ts
export class ColumnColorHelper {
  getColor(i: number): number {}
  getColors(): UInt32Array {}
}

// usage
col.meta.colors.grid = view.grid;
col.meta.colors.getColor(5);
col.meta.colors.getColors();
```
Please see the comments in https://github.com/datagrok-ai/public/issues/1125, I've added the `grid` property, which should be set to obtain the expected result from these methods. Alternatives are suggested in #1125 issue.

onuf:
Untied the grid dependency present in the previous implementation

PR #1203, opened 2022-11-16
GitHub Actions: CI: Remove always from packages workflow

dnillovna:


dnillovna:
@ksumag, take a look at the PR. I changed the docker push behavior

PR #1459, opened 2023-01-09
Oserhiienko/descriptors

Aleksashka11:


StLeonidas:
- Add parallelization (See how in reddata/python/grok_compute/utilities.py)
- Add progress indicator

Aleksashka11:
> * Add parallelization (See how in reddata/python/grok_compute/utilities.py)
Ok, I'll add
> * Add progress indicator
It is added, isn't it?
`const pi = DG.TaskBarProgressIndicator.create('Calculating descriptors...');`

packages/Chem/src/descriptors/descriptors-calculation.ts
```diff
@@ -14,12 +15,19 @@ export async function addDescriptors(smilesCol: DG.Column, viewTable: DG.DataFra
   openDescriptorsDialog(await getSelected(), async (selected: any) => {
     await grok.dapi.userDataStorage.postValue(_STORAGE_NAME, _KEY, JSON.stringify(selected));
     selected = await getSelected();
-    const pi = DG.TaskBarProgressIndicator.create('Calculating descriptors');
```
alex-aprm, line 17: Exception must be handled. I believe we need a generic method for this, such as safeCall in the core

packages/Chem/src/descriptors/descriptors-calculation.ts
```diff
@@ -14,12 +15,19 @@ export async function addDescriptors(smilesCol: DG.Column, viewTable: DG.DataFra
...
       DG.DataFrame.fromColumns([DG.Column.fromList('string', 'selected', selected)]),
-    );
+    ).then((res) => {
+      table = res;
+      pi.update(70, 'Calculating descriptors..: 70% completed');
+    });
+    pi.update(90, 'Adding results..: 90% completed');
     addResultColumns(table, viewTable);
+    pi.update(100, 'Descriptors are calculated');
     pi.close();
```
alex-aprm, line 31: Should be in finally block

packages/Chem/src/descriptors/descriptors-calculation.ts
```diff
@@ -14,12 +15,19 @@ export async function addDescriptors(smilesCol: DG.Column, viewTable: DG.DataFra
   openDescriptorsDialog(await getSelected(), async (selected: any) => {
```
alex-aprm, line 15: It should be pure function that aceepts column and returns columns, optionally providing progress information to caller

packages/Chem/src/descriptors/const.ts
```diff
@@ -0,0 +1,1321 @@
+export const descriptors: any = {
```
alex-aprm, line 1: Maybe, it's better to store it in JSON file, and not to build in to the package?

PR #1550, opened 2023-01-31
molecules should display also if they fail to kekulize

ptosco:
Please see the attached CSV file.
[fail_to_display_dg.csv](https://github.com/datagrok-ai/public/files/10548812/fail_to_display_dg.csv)

With the current `Chem` package the molecules in the `Core` column show up as a red cross as they fail to kekulize:
image
Also, an uncaught exception shows up in the console:
image
This should not happen - in such cases structures should be shown in their aromatic form.
This PR fixes the problem:
image


StLeonidas:
Closing PR due to duplicated functionality, though safeMol code was used.

PR #1554, opened 2023-02-01
Rename the property panel to context panel

onuf:


onuf:
@dnillovna please check the workflow run, there are errors with css files on Help / Convert to HTML step.

PR #1555, opened 2023-02-01
Add wizard hints

onuf:


skalkin (changes requested):
It's not clear at all what it is or how to use it. Let's make the API/documentation more clear.

js-api/ui.ts
```diff
@@ -1450,11 +1451,70 @@ export namespace hints {
...
+
+  /** Called when the page is activated */
+  onActivated?: () => void;
+
+  /** Returns error message (and stops wizard from proceeding to the next page),
+   * or null if validated */
+  validate?: () => string | null;
+
+  /** Displays the wizard next to the specified element when the page is current. */
+  showNextTo: HTMLElement;
```
onuf, line 1516: The very first idea was to simply extend the WizardPage interface, but I didn't manage to due to the 'root' property types incompatibility. The objective was to (1) make `showNextTo` obligatory in the hints API; (2) make the root optional (we provide a simpler way to set instructions via `text`) (decided not to introduce this change to the wizard itself, only in the hints API)

packages/ApiSamples/scripts/ux/Interactivity/hints.js
```diff
@@ -13,10 +13,23 @@ $(addNCIcon).on('click', () => {
 });
 
 grok.events.onDialogShown.pipe(
-  rxjs.operators.first(),
-  rxjs.operators.filter((dlg) => dlg.title === 'Add New Column'))
+  rxjs.operators.filter((dlg) => dlg.title === 'Add New Column'),
+  rxjs.operators.first())
```
onuf, line 17: Change the operators to make the subscription work (due to the wizard, 'Add New Column' dialog isn't the first one to open)

js-api/ui.ts
skalkin: Nope, WizardPage does not have to know about hints at all, it works the other way. There is no reason for that showNextTo field to exist, it only complicates the API and confuses the developer.

onuf: Fixed.

js-api/ui.ts
```diff
@@ -1450,11 +1451,74 @@ export namespace hints {
...
+   * from the dialog header (the "x" icon), via the "Cancel" button, or via the [Wizard.close()] method.
+   * Example: {@link https://public.datagrok.ai/js/samples/ux/Interactivity/hints}:*/
+  export function addTextHint(options: {title?: string, helpUrl?: string, pages?: HintPage[]}): Wizard {
+    let targetElement: HTMLElement | undefined;
+    const overlay = div([], 'ui-hint-overlay');
+    $('body').prepend(overlay);
+    const horizontalOffset = 10;
+    const wizard = new Wizard({title: options.title, helpUrl: options.helpUrl});
+    if (options.pages) {
+      options.pages.forEach((p) => {
```
vdyma, line 1467: Using `for` loop instead of `.forEach` is prefered.

js-api/ui.ts
```diff
@@ -1450,11 +1451,74 @@ export namespace hints {
...
+    const _x = $(wizard.root).css('left');
+    const _y = $(wizard.root).css('top');
+    wizard.show({width: 250, height: 200,
+      x: _x ? parseInt(_x) : undefined, y: _y ? parseInt(_y) : undefined});
+    return wizard;
+  }
+
   /** Removes the hint indication from the provided element and returns it. */
-  export function remove(el: HTMLElement): HTMLElement {
+  export function remove(el?: HTMLElement): HTMLElement | null {
```
vdyma, line 1495: It's better if the function returns `null` immediately when `el` is `undefined`

js-api/ui.ts
```diff
@@ -1450,11 +1451,74 @@ export namespace hints {
...
+  root?: HTMLDivElement;
+
+  /** Caption to be displayed on top of the panel */
+  caption?: HTMLElement | string;
+
+  /** Adds a div with the specified text as the page root. */
+  text?: string;
+
+  /** Displays the wizard next to the specified element when the page is current. */
+  showNextTo: HTMLElement;
```
vdyma, line 1516: Can this property be `undefined`?

js-api/src/ui/wizard.ts
```diff
@@ -32,7 +32,7 @@ export class Wizard extends Dialog {
   nextButton = this.addButton('>>', () => this.next());
 
   constructor(options?: {title?: string, helpUrl?: string}) {
-    super(api.grok_Dialog(options?.title, options?.helpUrl, true, true));
+    super(api.grok_Dialog(options?.title, options?.helpUrl, !!(options?.title || options?.helpUrl), true));
```
dionissqq, line 35: we need two of them !! ?

onuf: Here it's simply a shorter way to convert to boolean

js-api/ui.ts
onuf: No, this one is required, it would be logical to move it above the optional properties.

Issue #1571, opened 2023-02-03
Charts: GlobeViewer tests fail


onuf:
Test result : Failed : 3139 : Charts.Viewers.GlobeViewer : Error: cannot load viewer

LaykDimon:
done

Issue #1570, opened 2023-02-03
Charts: SankeyViewer tests fail


onuf:
_Test result : Failed : 3526 : Charts.Viewers.SankeyViewer : Error: cannot load viewer from layout_
_Charts 1.0.16_ 
Sankey Viewer does not work with empty data

![Recording 2023-02-06 at 16 38 15](https://user-images.githubusercontent.com/70022930/217000653-fc3a6569-6846-4862-94a1-056355e982bf.gif)

LaykDimon:
done

PR #1586, opened 2023-02-09
dkovalyov/Charts-additions

LaykDimon:
Demos were added for all the existing viewers.
Also, the sunburst viewer was refactored.
Issues #1570 and #1571 were fixed.


packages/Charts/src/demos/demo-data.ts
```diff
@@ -0,0 +1,263 @@
+import * as DG from 'datagrok-api/dg';
+
+export const energyUK = DG.DataFrame.fromCsv(
```
onuf, line 3: Why do you use this way of storing demo tables? For data that isn't present in the demo files connection, there's an option to use [package files API](https://github.com/datagrok-ai/public/blob/master/help/develop/how-to/access-data.md#package-files). It seems that the tables are mostly from the demo connection, so you can simply open it via API, e.g.: `grok.data.getDemoTable('geo/earthquakes.csv')`

onuf: The same goes for the test data. The package is big enough already, so unless there's a reason to add data this way, I don't see why we need these files here.

LaykDimon: thank you, guess you are right, didn't think about `grok.data.getDemoTable()`
will fix it

LaykDimon:
unnecessary demo-data.ts and test-data.ts files were removed
also, utils.ts in tests were moved to separate folder

packages/Charts/src/demos/demo.ts
```diff
@@ -0,0 +1,73 @@
...
+    tableView.addViewer('ChordViewer');
+  }
+
+  static async globeViewerDemo() {
+    const df = await grok.data.getDemoTable('geo/earthquakes.csv');
+    const tableView = grok.shell.addTableView(df);
+    // timeout is for the table view to fully load, activate detectors and then start the globe viewer
+    setTimeout(() => {
+      tableView.addViewer('GlobeViewer');
+    }, 300);
```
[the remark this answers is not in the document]
LaykDimon: for sure, i was looking for something like that, thanks!

packages/Charts/src/demos/demo.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  static timelinesViewerDemo() {
+    const df = DG.DataFrame.fromCsv(
+      `USUBJID, AESTDY, AEENDY, SEX, AGE
+      s1, 10/02/2018, 10/09/2018, F, 48
+      s2, 10/04/2018, 10/07/2018, M, 51
+      s3, 10/02/2018, 10/05/2018, F, 39
+      s4, 10/07/2018, 10/08/2018, M, 43`);
+    const tableView = grok.shell.addTableView(df);
+    tableView.addViewer('TimelinesViewer');
```
onuf, line 59: A small suggestion: you can use the docking API to place a viewer in a more suitable position, for example:
```js
const viewer = tableView.addViewer('TimelinesViewer');
tableView.dockManager.dock(viewer, DG.DOCK_TYPE.DOWN);
```
It's up to you to decide on the most suitable position, I used this particular case because the demo data is small and we have an empty space under the table anyway

LaykDimon: hm, will think about a suitable position for the viewers in the demos, it may play a role in user perception

LaykDimon:
subscribe on semantic type detection was added in the globe viewer demo, but for now, it isn't used because of a bug, will be changed later
about docking API - went through all the demo charts and decided that it isn't needed there, default positioning works fine

LaykDimon:
replaced setTimeout in the globe demo with a DG.debounce

packages/Charts/src/demos/demo.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  static async sunburstViewerDemo() {
+    const df = await grok.data.getDemoTable('demog.csv');
+    const tableView = grok.shell.addTableView(df);
+    tableView.addViewer('SunburstViewer');
+  }
+
+  static async surfacePlotDemo() {
+    const df = await grok.data.getDemoTable('demog.csv');
```
skalkin, line 44: This dataset makes no sense for the surface plot. I suggest to programmatically create a good one, such as 3d parabolic function or something fancy like that. Perhaps @AnnaMuza already has a good case for that.

packages/Charts/src/demos/demo.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  static async globeViewerDemo() {
+    const df = await grok.data.getDemoTable('geo/earthquakes.csv');
+    const tableView = grok.shell.addTableView(df);
+    // semantic type detection is needed for the latitude, longitude and magnitude columns
+    DG.debounce(df.onSemanticTypeDetected, 50).subscribe((_) => tableView.addViewer('GlobeViewer'));
+  }
+
+  static async groupAnalysisViewerDemo() {
+    const df = await grok.data.getDemoTable('demog.csv');
```
skalkin, line 20: Need a better dataset (and maybe better default settings) for that. Perhaps @MariaDolotova might help.

packages/Charts/src/demos/demo.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  static timelinesViewerDemo() {
+    const df = DG.DataFrame.fromCsv(
+      `USUBJID, AESTDY, AEENDY, SEX, AGE
+      s1, 10/02/2018, 10/09/2018, F, 48
+      s2, 10/04/2018, 10/07/2018, M, 51
+      s3, 10/02/2018, 10/05/2018, F, 39
+      s4, 10/07/2018, 10/08/2018, M, 43`);
+    const tableView = grok.shell.addTableView(df);
+    tableView.addViewer('TimelinesViewer');
```
skalkin, line 57: @onuf , please come up with a better example that would showcase most of the timelines functionality

packages/Charts/src/package.ts
```diff
@@ -12,123 +13,158 @@ import {GroupAnalysisViewer} from './viewers/group-analysis/group-analysis-viewe
...
 export function _radarViewerDemo() {
-  radarViewerDemo();
+  Demo.radarViewerDemo();
 }
 
-//name: radarViewerDemo
+//name: sankeyViewerDemo
+//meta.demoPath: Viewers | Sankey
+export function _sankeyViewerDemo() {
+  Demo.sankeyViewerDemo();
```
skalkin, line 139: Since all the demos are so similar at the moment, would it make sense to simply make one parameterized function viewerDemo(table, options) and simply use it directly here?

skalkin (commented):
Great start and you are free to merge it yourself, but let's make sure to address the comments I've made

packages/Charts/src/package.ts
LaykDimon: it makes sense to make one parameterized function, yes
but I guess it's better to use a function like `viewerDemo(viewerName, options)` to transfer the viewer name and options for it
I don't really think that we need the possibility to choose a table in the viewer demo function as the program will choose the appropriate table for the chart by itself

packages/Charts/src/demos/demo.ts
LaykDimon: done

packages/Charts/src/demos/demo.ts
LaykDimon: done

packages/Charts/src/package.ts
LaykDimon: guess it's done

packages/Charts/src/demos/demo.ts
LaykDimon: done

Issue #1353, opened 2022-12-15
Tutorials: increase test coverage


onuf:
- [ ] add checks for tutorial prerequisites (done for packages, the same is expected for CVM, db connections, etc.)

dnillovna:
This issue has been mirrored in Jira: https://reddata.atlassian.net/browse/GROK-14843

dnillovna:
This issue has been mirrored in Jira: https://reddata.atlassian.net/browse/GROK-16097

PR #1602, opened 2023-02-16
#1353 Tutorials: add prerequisites annotation for tutorials

onuf:


onuf:
I'll probably add information on the modeling component to `grok.dapi.admin.getServiceInfos` as there's no information on model engines, such as H2O, currently. This will be useful in "Predictive Modeling" and "Virtual Screening" tutorials.

Issue #1524, opened 2023-01-25
Bio: PepSeA enhancements


vdyma:
PepSeA tool can be improved in several ways:

- [x] automatically generate result column name
- [x] create unit tests and benchmarks
- [x] add parameters hints
- [x] merge with MSA panel and choose the tool based on data
- [ ] add to documentation
- [x] make Clusters field nullable

PR #1603, opened 2023-02-16
#1524: MSA by clusters

vdyma:


packages/Bio/src/package.ts
```diff
@@ -423,17 +423,18 @@ export function multipleSequenceAlignmentAny(col: DG.Column<string> | null = nul
...
-      performAlignment = () => runKalign(potentialCol, false, unUsedName);
-    } else if (checkInputColumnUi(potentialCol, 'MSA', [NOTATION.HELM], [], false)) {
+      performAlignment = () => runKalign(potentialCol, false, unUsedName, clustersColInput.value);
+    } else if (checkInputColumnUi(potentialCol, potentialCol.name, [NOTATION.HELM], [], false)) {
       for (const inputRootStyle of inputRootStyles)
         inputRootStyle.display = 'initial';
 
-      performAlignment = () => runPepsea(
-        potentialCol, unUsedName, methodInput.value!, gapOpenInput.value!, gapExtendInput.value!);
+      performAlignment = () => runPepsea(potentialCol, unUsedName, methodInput.value!, gapOpenInput.value!,
```
LaykDimon, line 436: maybe unusedName, not unUsedName, don't think we need to make an accent on ...Used...

packages/Bio/src/utils/multiple-sequence-alignment.ts
```diff
@@ -11,6 +11,8 @@ import Aioli from '@biowasm/aioli';
 import {AlignedSequenceEncoder} from '@datagrok-libraries/bio/src/sequence-encoder';
 import {UnitsHandler} from '@datagrok-libraries/bio/src/utils/units-handler';
 
+type KalignBodyUnit = {};
```
LaykDimon, line 14: why do we need this empty type?

packages/Bio/src/utils/multiple-sequence-alignment.ts
```diff
@@ -29,31 +31,55 @@ function _stringsToFasta(sequences: string[]): string {
...
+    clustersCol = clustersCol.convertTo(DG.TYPE.STRING);
+  clustersCol.compact();
+  
+  const clustersColCategories = clustersCol.categories;
+  const clustersColData = clustersCol.getRawData();
+  const fastaSequences: string[][] = new Array(clustersColCategories.length);
+  const clusterIndexes: number[][] = new Array(clustersColCategories.length);
+  for (let rowIdx = 0; rowIdx < sequencesLength; ++rowIdx) {
+    const clusterCategoryIdx = clustersColData[rowIdx];
+    (fastaSequences[clusterCategoryIdx] ??= []).push(sequences[rowIdx]);
```
LaykDimon, line 53: can we use another for and access by index instead of push?

packages/Bio/src/utils/multiple-sequence-alignment.ts
```diff
@@ -29,31 +31,55 @@ function _stringsToFasta(sequences: string[]): string {
...
+    const clusterSequences = fastaSequences[clusterIdx];
+    const fasta = _stringsToFasta(clusterSequences);
+    
+    console.log(['fasta.length =', fasta.length]);
 
-  await CLI.fs.writeFile('input.fa', fasta);
-  const output = await CLI.exec('kalign input.fa -f fasta -o result.fasta');
-  console.warn(output);
+    await CLI.fs.writeFile('input.fa', fasta);
+    const output = await CLI.exec('kalign input.fa -f fasta -o result.fasta');
```
LaykDimon, line 70: maybe put this command into const?

packages/Bio/src/utils/multiple-sequence-alignment.ts
```diff
@@ -29,31 +31,55 @@ function _stringsToFasta(sequences: string[]): string {
...
  * @return {Promise<DG.Column>} Aligned sequences.
  */
-export async function runKalign(srcCol: DG.Column, isAligned = false, unUsedName: string = ''): Promise<DG.Column> {
-  let sequences = srcCol.toList();
+export async function runKalign(srcCol: DG.Column<string>, isAligned: boolean = false, unUsedName: string = '',
+  clustersCol: DG.Column | null = null): Promise<DG.Column> {
+  let sequences: string[] = srcCol.toList();
 
   if (isAligned)
     sequences = sequences.map((v: string, _) => AlignedSequenceEncoder.clean(v).replace(/\-/g, ''));
```
onuf, line 39: Why do we need the second argument in the `map` callback (`_`)?

packages/Bio/src/utils/multiple-sequence-alignment.ts
vdyma: No, but it's better to use formatted string for file names

PR #1614, opened 2023-02-24
dkovalyov/fit-curve-renderer

LaykDimon:
It's not the latest version and the code hasn't been well-reviewed by me
nevertheless would be grateful to listen to advices


packages/PowerGrid/src/fit/fit-curve.ts
```diff
@@ -0,0 +1,240 @@
...
+  for (let i = 0; i < coordinates.x.length; i++) {
+    scaledCoordinates.x[i] = canvasXCenter + coeff * (coordinates.x[i] - xMid);
+    scaledCoordinates.y[i] = canvasYCenter + coeff * (coordinates.y[i] - yMid);
+  }
+
+  return scaledCoordinates;
+}
+
+export function getParams(columns: DG.Column[]): number[] {
+  const coordinates: {x: number[], y: number[]} = {x: columns[0].toList(), y: columns[1].toList()};
```
vdyma, line 39: Consider using `columns[0].getRawData()` as it will be memory-efficient

vdyma: Do we need coordinates object at all? We could just use two variables.

packages/PowerGrid/src/fit/fit-curve.ts
```diff
@@ -0,0 +1,240 @@
...
+
+  onClick(gridCell: DG.GridCell, e: MouseEvent): void {
+    const df = gridCell.grid.dataFrame;
+
+    const dfColumn = df.columns.toList().filter((column) => column.type === DG.COLUMN_TYPE.DATA_FRAME)![0];
+    const coordinateDf = DG.toJs(dfColumn.get(gridCell.cell.row.idx)) as DG.DataFrame;
+    if (!coordinateDf) return;
+
+    const coordinateColumns = coordinateDf.columns.byNames(['x', 'y']);
+    const coordinates: {x: number[], y: number[]} = {x: coordinateColumns[0].toList(),
```
vdyma, line 85: Consider using `columns[0].getRawData()` as it will be memory-efficient

vdyma: Do we need coordinates object at all? We could just use two variables.

packages/PowerGrid/src/fit/fit-curve.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    const df = gridCell.grid.dataFrame;
+    if (w < 20 || h < 10 || df === void 0) return;
+
+    const dfColumn = df.columns.toList().filter((column) => column.type === DG.COLUMN_TYPE.DATA_FRAME)![0];
+    const coordinateDf = DG.toJs(dfColumn.get(gridCell.cell.row.idx)) as DG.DataFrame;
+    if (!coordinateDf) return;
+
+    const coordinateColumns = coordinateDf.columns.byNames(['x', 'y']);
+    const coordinates: {x: number[], y: number[]} = {x: coordinateColumns[0].toList(),
```
vdyma, line 146: Getting coordinates seem to be a common operation for this renderer. Consider creating a separate function for that.

packages/PowerGrid/src/fit/fit-curve.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    const updatedWidth = 160;
+    gridCell.gridColumn.width = updatedWidth;
+
+    const canvasPointCoordinates = scaleCoordinates(coordinates, gridCell.gridColumn.width, h);
+
+    const filteredCoordinateDf = coordinateDf
+      .groupBy(['x', 'y'])
+      .whereRowMask(coordinateDf.filter)
+      .aggregate();
```
vdyma, line 157: Basically you're just copying the dataframe, right? Consider using `coordinateDf.clone(coordinateDf.filter, ['x', 'y'])`

packages/PowerGrid/src/fit/fit-curve.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    const filteredColumns = filteredCoordinateDf.columns.byNames(['x', 'y']);
+    const filteredCoordinates: {x: number[], y: number[]} = {x: filteredColumns[0].toList(),
+      y: filteredColumns[1].toList()};
```
vdyma, line 161: Same as comments above

packages/PowerGrid/src/fit/fit-curve.ts
```diff
@@ -0,0 +1,240 @@
...
+      g.moveTo(x + canvasFitCurveCoordinates.x[0], (y + h) - canvasFitCurveCoordinates.y[0]);
+      for (let i = 1; i < canvasFitCurveCoordinates.x.length; i++)
+        g.lineTo(x + canvasFitCurveCoordinates.x[i], (y + h) - canvasFitCurveCoordinates.y[i]);
+      g.stroke();
+    }
+
+    if (data.seriesOptions?.showPoints) {
+      const filteredIndexes = coordinateDf.filter.getSelectedIndexes();
+      for (let i = 0; i < canvasPointCoordinates.x.length; i++) {
+        const color = DG.Color.scatterPlotMarker;
```
vdyma, line 199: This object recreates each iteration. Consider moving out of the loop

packages/PowerGrid/src/fit/fit-data.ts
```diff
@@ -0,0 +1,301 @@
...
+    if (p.defaultValue != null)
+      o[p.name] = p.defaultValue;
+  return o;
+}
+
+
+/** Merges properties of the two objects by iterating over the specified {@link properties}
+ * and assigning properties from {@link source} to {@link target} only when
+ * the property is not defined in target and is defined in source. */
+function mergeProperties(properties: Property[], source: any, target: any) {
```
vdyma, line 145: Avoid using any as input types

packages/PowerGrid/src/fit/fit-data.ts
```diff
@@ -0,0 +1,301 @@
...
+
+
+/** Returns existing, or creates new column default chart options. */
+export function getColumnChartOptions(gridColumn: DG.GridColumn): IFitChartData {
+  return gridColumn.tags[TAG_FIT] ??= createDefaultChartData();
+}
+
+
+export function getChartBounds(chartData: IFitChartData): DG.Rect {
+  let o = chartData.chartOptions;
```
vdyma, line 170: `o` can be `const`

packages/PowerGrid/src/fit/fit-data.ts
```diff
@@ -0,0 +1,301 @@
...
+  if (o?.minX && o.minY && o.maxX && o.maxY)
+    return new DG.Rect(o.minX, o.minY, o.maxX - o.minX, o.maxY - o.minY);
+  if (!chartData.series?.length || chartData.series.length == 0)
+    return new DG.Rect(0, 0, 1, 1);
+  else {
+    let bounds = getDataBounds(chartData.series[0].points);
```
vdyma, line 176: `bounds` can be `const`

packages/PowerGrid/src/fit/fit-data.ts
```diff
@@ -0,0 +1,301 @@
...
+  for (let i = 0; i < series.points.length; i++) {
+    if (series.points[i].y === medY) {
+      xAtMedY = series.points[i].x;
+      break;
+    }
+  }
+  let initialParams = [dataBounds.bottom, 1.2, xAtMedY, dataBounds.top];
+
+  return fit(
+    {x: series.points.map(p => p.x), y: series.points.map(p => p.x)},
```
vdyma, line 237: `y` is set the same as `x`. Is it intended?

packages/PowerGrid/src/fit/fit-renderer.ts
```diff
@@ -0,0 +1,84 @@
...
+    g.beginPath();
+    g.rect(x, y, w, h);
+    g.clip();
+
+    if (w < 20 || h < 10) return;
+    const screenBounds = new DG.Rect(x, y, w, h).inflate(-6, -6);
+
+    const data = getChartData(gridCell);
+    let dataBounds = getChartBounds(data);
+    let transform = Transform.linear(dataBounds, screenBounds);
```
vdyma, line 51: These could be constans

packages/PowerGrid/src/fit/fit-renderer.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    for (let series of data.series!) {
+      if (series.showPoints ?? true) {
```
vdyma, line 54: Consider trying to terminate as early as possible. Here it could be: `if (!series.showPoints) continue;`

packages/PowerGrid/src/fit/fit-curve.ts
LaykDimon: > Do we need coordinates object at all? We could just use two variables.

guess you're right

packages/PowerGrid/src/fit/fit-curve.ts
LaykDimon: > Do we need coordinates object at all? We could just use two variables.

guess you're right

packages/PowerGrid/src/fit/fit-curve.ts
LaykDimon: don't even need cloning or aggregating the dataframe, can just go through df.filter.getSelectedIndexes()

packages/PowerGrid/src/fit/fit-data.ts
LaykDimon: i'll look at it

js-api/src/grid.ts
```diff
@@ -13,8 +13,13 @@ import {IFormLookSettings, IGridLookSettings} from "./interfaces/d4";
 let api = <any>window;
 let _bytes = new Float64Array(4);
 
+export interface IPoint {
```
dionissqq, line 16: do we really need it?

packages/PowerGrid/src/fit/fit-curve.ts
```diff
@@ -0,0 +1,241 @@
...
+  const xMid = (xMax + xMin) / 2;
+  const yMid = (yMax + yMin) / 2;
+
+  const coeff = xDiff * canvasHeight >= yDiff * canvasWidth ? canvasWidth / xDiff : canvasHeight / yDiff;
+  const canvasXCenter = canvasWidth / 2;
+  const canvasYCenter = canvasHeight / 2;
+
+  const scaledCoordinates: {x: number[], y: number[]} = {x: [], y: []};
+  for (let i = 0; i < coordinates.x.length; i++) {
+    scaledCoordinates.x[i] = canvasXCenter + coeff * (coordinates.x[i] - xMid);
```
dionissqq, line 31: same as  = coef * (coordinates.x[i] - xMin)?

packages/PowerGrid/src/fit/fit-curve.ts
```diff
@@ -0,0 +1,241 @@
...
+    scaledCoordinates.y[i] = canvasYCenter + coeff * (coordinates.y[i] - yMid);
+  }
+
+  return scaledCoordinates;
+}
```
dionissqq, line 36: about this function scaleCoordinates
not sure this is right approach
probably best would be if we stored only native data and all other stuff like mapping to screen would be done separately by transformation (methods xToScreen, yToScreen, screenToX, screenToY, hitTest)
Check out how it is done for scatterplot and other viewers
Thoug for now lets not change it

PR #1620, opened 2023-02-28
Vmakarichev/datagrok with cpp

vmakarichev:
Runtime system for wasm-functions call in webworker.


packages/vmakarichev-datagrok_with_cpp/CallWasmWW/wasm/callWasmUtils.js
```diff
@@ -0,0 +1,710 @@
...
+                let array;
+
+                // this is OK if type of column and target type coinside
+                //array = inputVals[i].getRawData();
+
+                let col = inputVals[i];
+
+                // here, we check types and perform an appropriate transform
+                if(((col.type == INT_TYPE) && (type == INT_COLUMN_TYPE)) 
+                   || ((col.type == DOUBLE_TYPE) && (type == FLOAT_COLUMN_TYPE)))
```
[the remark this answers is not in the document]
LaykDimon: also space before the brackets

packages/vmakarichev-datagrok_with_cpp/CallWasmWW/wasm/callWasmUtils.js
```diff
@@ -0,0 +1,710 @@
...
+    //console.log('inputs after memory allocation:');
+    //console.log(inputs);
+} // allocateMemoryForBuffer
+
+// Get array of values that are put to wasm-function.
+function getArrOfWasmParams(inputs) {
+    let params = [];
+
+    // Process each type of input
+    for(let arg of inputs) {
```
LaykDimon, line 269: consider changing let to const if the value isn't changed (like arg here)

LaykDimon: same example everywhere in the code

packages/vmakarichev-datagrok_with_cpp/CallWasmWW/wasm/callWasmUtils.js
```diff
@@ -0,0 +1,710 @@
...
+                break;
+            
+            // columns & new columns
+            case INT_COLUMNS_TYPE:
+            case NEW_INT_COLUMNS_TYPE:
+            case FLOAT_COLUMNS_TYPE:
+            case NEW_FLOAT_COLUMNS_TYPE:
+                types.push(NUMBER);
+                types.push(NUMBER);
+                types.push(NUMBER);
```
LaykDimon, line 337: guess you've created `shiftMap` for these purposes, so you can use the length from there and use it here in the for cycle?

packages/vmakarichev-datagrok_with_cpp/CallWasmWW/wasm/callWasmUtils.js
```diff
@@ -0,0 +1,710 @@
...
+    return types;
+} // getArrOfWasmTypes
+
+// Put array data to buffer
+function putDataToBuffer(module, inputs) {
+    let shift;
+    let heap;    
+    
+    for(let arg of inputs) {
+        let type = arg.type;
```
LaykDimon, line 355: again, use const for `arg` and `type`

packages/vmakarichev-datagrok_with_cpp/CallWasmWW/wasm/callWasmUtils.js
```diff
@@ -0,0 +1,710 @@
...
+        if(key == CALL_RESULT)             
+            continue;
+
+        // here, we consider each type of input
+        switch(type){           
+
+            // numbers
+            case NUM_TYPE:                
+            case INT_TYPE:
+            case DOUBLE_TYPE:
```
LaykDimon, line 97: i have a suggestion to repalce the switch code with object literals
it's more maintainable and easy-extendable
something like
```
const typeDefinition = {
        'NUM_TYPE': () => {
          arg.data = inputVals[i];
          i++;
        },
        'FLOAT_COLUMN_TYPE': () => {
          console.log('...');
        },
        'default': () => console.log('...')
      };
```
and so on

LaykDimon: if there are similar functions for different objects (like NUM_TYPE, INT_TYPE, etc.), you can just write a function for it once and then reuse it in this mapping

LaykDimon: and then you just do something like `typeDefinition[type]`

LaykDimon: this will be better just because switches are hardly expandable and also always require a break everywhere, which is a bit confusing

packages/vmakarichev-datagrok_with_cpp/CallWasmWW/wasm/callWasmUtils.js
```diff
@@ -0,0 +1,710 @@
...
+    let ref;
+
+    // complete an input for cpp
+    let i = 0;
+    for(let key in argsSpecification){
+        let arg = argsSpecification[key]; 
+        let type = arg.type;
+        
+        // skip auxiliry element
+        if(key == CALL_RESULT)
```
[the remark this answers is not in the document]
vmakarichev: Agree

packages/vmakarichev-datagrok_with_cpp/CallWasmWW/wasm/callWasmUtils.js
vmakarichev: Thanks! I'll consider this

packages/vmakarichev-datagrok_with_cpp/CallWasmWW/wasm/callWasmUtils.js
vmakarichev: Thanks! I'll consider this

PR #1636, opened 2023-03-08
Vmakarichev/web autosolver

vmakarichev:
This package provides solution of initial value problems (IVPs).
The following routine steps are automated:
text description of IVP -> C++-code -> wasm-code -> package function (with RichFunction view annotation).
Text description of the problem is defined as a set of lines (see, for example, https://github.com/datagrok-ai/public/blob/vmakarichev/WebAutosolver/packages/WebAutosolver/scripts/MarkovModel.txt).
Then, the script create.py should be launched. After that, the package is ready for publishing.
Currently, the problem FAE is solved.


packages/WebAutosolver/scripts/export.py
```diff
@@ -0,0 +1,685 @@
...
+        else:
+            self.processOutputLineWithoutNew(line)
+
+    def processInputLine(self, line):
+        """ Process '//input'-line with arguments that are passed.
+        """
+        self.annotation.append(line)
+
+        # get type and name part of line
+        ignore1, ignore2, rest = line.partition(': ')
```
LaykDimon, line 92: magic strings - `: `, `, ' - you're using the multiple times

PR #1638, opened 2023-03-09
dkovalyov/controls

LaykDimon:
providing typeAhead, dropDown and breadcrumbs controls for the platform


libraries/utils/src/shared-components/drop-down.ts
```diff
@@ -1,66 +1,91 @@
...
+    this._dropDownElement.addEventListener('mouseleave', () => {
       this._isMouseOverElement = false;
     }, false);
-  }
 
-  private _updateElements(elements: string[]) {
-    this._elements = elements;
+    document.addEventListener('click', (event) => {
+      if (this.root.contains(event.target as Node)) return;
+      if (!this.isExpanded) return;
```
vdyma, line 60: This could be one if with a union of two statements

libraries/utils/src/shared-components/drop-down.ts
```diff
@@ -1,66 +1,91 @@
...
 
 
-  get root() {
-    return this._root!;
+  get onExpand(): Observable<MouseEvent> {
+    return fromEvent<MouseEvent>(this.root, 'click')
+      .pipe(
+        filter(() => !this._isMouseOverElement)
+      );
+  }
```
vdyma, line 84: Such methods separation (each call on new line) is redundant. Consider the following:
```
get onExpand(): Observable<MouseEvent> {
  return fromEvent<MouseEvent>(this.root, 'click').pipe(filter(() => !this._isMouseOverElement));
}
```

libraries/utils/css/breadcrumbs.css
```diff
@@ -0,0 +1,24 @@
+.breadcrumbs {
```
onuf, line 1: More of a discussion opener than a direct suggestion: isn't the class name too generic? We typically use prefixes, such as `ui`, `d4`, or shortened package names.

libraries/utils/src/shared-components/drop-down.ts
```diff
@@ -1,66 +1,91 @@
...
-  private _expanded: boolean;
+  private _label: string | Element;
+
+  isExpanded: boolean;
+  root: HTMLDivElement;
 
-  constructor(elements: string[]) {
-    this._elements = [];
+
+  constructor(label: string | Element, elementCreation: () => HTMLElement) {
```
onuf, line 18: Typically, verbs are used in function names (`getElement` or `createElement` instead of `elementCreation`)

PR #1646, opened 2023-03-12
GROK-12450: Viewers | Filters: Custom filter test

tanas80:
There was a problem to display the custom PTM filter in MLB.

The problem was solved by specifying a column name to associate the custom filter with, though the PTM filter works on data of multiple columns.

Tests for the ApiTests package have been created to test displaying custom filters, they also are examples showing how to use custom filters.

### DIFF
--- a/packages/ApiTests/src/viewers/filters.ts
+++ b/packages/ApiTests/src/viewers/filters.ts
@@ -1,0 +1,74 @@
+import * as ui from 'datagrok-api/ui';
+import * as grok from 'datagrok-api/grok';
+import * as DG from 'datagrok-api/dg';
+
+import {
+  after,
+  before,
+  category,
+  expect,
+  test,
+  delay,
+  awaitCheck,
+  expectArray
+} from '@datagrok-libraries/utils/src/test';
+import {TableView} from 'datagrok-api/dg';
+import {_package} from '../package-test';
+
+import $ from 'cash-dom';
+
+category('Viewers: Filters', () => {
+  const csv1: string = `id1,id2,id3
+id1_0001,id2_001,id3_1
+id1_0002,id2_002,id3_2
+id1_0003,id2_003,id3_3`;
+
+  test('twoCategorical', async () => {
+    const df: DG.DataFrame = DG.DataFrame.fromCsv(csv1);
+    const view: TableView = grok.shell.addTableView(df);
+
+    const filterList: { [p: string]: string }[] = [
+      {column: 'id1', type: DG.FILTER_TYPE.CATEGORICAL, label: 'id1 label'},
+      {column: 'id2', type: DG.FILTER_TYPE.CATEGORICAL, label: 'id2 label'},
+    ];
+    const filtersViewer = view.filters({filters: filterList}) as DG.Viewer;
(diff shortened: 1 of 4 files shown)


packages/ApiTests/src/viewers/filters.ts
```diff
@@ -0,0 +1,74 @@
+import * as ui from 'datagrok-api/ui';
+import * as grok from 'datagrok-api/grok';
+import * as DG from 'datagrok-api/dg';
+
+import {
```
AnnaMuza, line 5: some unnecessary imports

packages/ApiTests/src/viewers/filters.ts
```diff
@@ -0,0 +1,74 @@
...
+  test('twoCategorical', async () => {
+    const df: DG.DataFrame = DG.DataFrame.fromCsv(csv1);
+    const view: TableView = grok.shell.addTableView(df);
+
+    const filterList: { [p: string]: string }[] = [
+      {column: 'id1', type: DG.FILTER_TYPE.CATEGORICAL, label: 'id1 label'},
+      {column: 'id2', type: DG.FILTER_TYPE.CATEGORICAL, label: 'id2 label'},
+    ];
+    const filtersViewer = view.filters({filters: filterList}) as DG.Viewer;
+    const filtersViewerDn: DG.DockNode = view.dockManager.dock(
```
AnnaMuza, line 35: what is the filtersViewerDn variable for?

packages/ApiTests/src/viewers/filters.ts
```diff
@@ -0,0 +1,74 @@
...
+      filtersViewer, DG.DOCK_TYPE.LEFT, null, 'Filters', 0.4);
+
+    await awaitCheck(() => {
+      let ok: boolean = true;
```
AnnaMuza, line 39: maybe ok var is also unnecessary and we can just return true?

packages/ApiTests/src/viewers/test-custom-filter.ts
```diff
@@ -0,0 +1,30 @@
...
+      delete state.column;
+      state.columnName = columnName;
+    }
+    super.applyState(state);
+  }
+
+  // -- Filter --
+
+  applyFilter(): void {
+    const k = 11;
```
AnnaMuza, line 28: leave it empty?

packages/ApiTests/src/viewers/filters.ts
tanas80: I import namespaces grok, DG and ui in all modules.
https://datagrok.slack.com/archives/C02AJ2FT2F5/p1679389910994899

packages/ApiTests/src/viewers/filters.ts
tanas80: removed

packages/ApiTests/src/viewers/filters.ts
tanas80: removed

packages/ApiTests/src/viewers/test-custom-filter.ts
tanas80: this was for the breakpoint for debugging, removed

PR #1760, opened 2023-04-04
GROK-12702: JS API: Added TagsInput control

LaykDimon:
Implementing the TagsInput control


libraries/utils/src/shared-components/tags-input.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    this._initEventListeners();
+  }
+
+	_initEventListeners() {
+		this.input.addEventListener('keyup', (event: KeyboardEvent) => {
+      if (event.code === 'Enter')
+        this.addTag((this.input as HTMLInputElement).value);
+    });
+	}
```
vdyma, line 32: Something is wrong with the formatting of the function `_initEventListeners`

libraries/utils/src/shared-components/tags-input.ts
```diff
@@ -0,0 +1,52 @@
...
+
+
+  addTag(tag: string) {
+    if (!this.tags.includes(tag)) {
```
vdyma, line 36: I feel like a better style is to terminate ASAP, so this could be:
```
if (this.tags.includes(tag))
  return;
// rest of the code here
```

libraries/utils/src/shared-components/tags-input.ts
```diff
@@ -0,0 +1,52 @@
...
+export class TagsInput extends DG.InputBase {
+  tags: string[];
+  _tagsDiv: HTMLDivElement;
+
+  constructor(tags: string[]) {
+    const inputElement = ui.stringInput('', '');
+    super(inputElement.dart);
+
+    this.tags = tags;
+    this._tagsDiv = ui.div(tags.map((tag) => ui.span([ui.span([tag])], `ui-tag ${tag}`)), 'ui-tag-list');
```
alex-aprm, line 21: Don't let unsanitized values go to css attribute. use data-tag attribute to put tag to DOM

libraries/utils/src/shared-components/tags-input.ts
LaykDimon: I still can't understand why it's displayed like that, because I don't have wrong tabulation locally and it is pushed

libraries/utils/src/shared-components/tags-input.ts
```diff
@@ -0,0 +1,154 @@
...
+    this._onTagRemoved.next(tag);
+  }
+
+  getTags(): string[] {
+    return this._tags;
+  }
+
+  setTags(tags: string[]) {
+    this._tags = tags;
+    this._tagsDiv = ui.div(tags.map((tag) => { return this._createTag(tag); }), 'ui-tag-list');
```
onuf, line 141: As a suggestion, since these lines repeat several times, you can create a helper method with them:
```ts
this._tags = tags;
this._tagsDiv = ui.div(tags.map((tag) => { return this._createTag(tag); }), 'ui-tag-list');
```

LaykDimon: sure

PR #1766, opened 2023-04-07
Compute: Customization support for pipeline view p1

andrey-sharapov-sc:
- allow function views to accept DG.FuncCall
- reimplement onFuncCallReady as a method
- use funcCallReplaced as a single before draw event API
- proper filtering of step calls via ids
- add some expiremental hooks in pipeline view

### DIFF
--- a/libraries/compute-utils/function-views/src/function-view.ts
+++ b/libraries/compute-utils/function-views/src/function-view.ts
@@ -14,7 +14,6 @@ export abstract class FunctionView extends DG.ViewBase {
   protected _funcCall?: DG.FuncCall;
   protected _lastCall?: DG.FuncCall;
   protected _type: string = 'function';
-  protected funcName: string;
 
   // emitted when after a new FuncCall is linked
   public funcCallReplaced = new Subject<true>();
@@ -32,7 +31,6 @@ export abstract class FunctionView extends DG.ViewBase {
   ) {
     super();
     this.box = true;
-    this.funcName = initValue instanceof DG.FuncCall ? initValue.func.name : initValue;
     this.init();
   }
 


libraries/compute-utils/function-views/src/function-view.ts
```diff
@@ -14,46 +14,46 @@ export abstract class FunctionView extends DG.ViewBase {
...
    * @param options Configuration object for the view.
    */
   constructor(
-    protected funcName: string,
+    protected initValue: string | DG.FuncCall,
     public options: {historyEnabled: boolean, isTabbed: boolean} = {historyEnabled: true, isTabbed: false},
   ) {
     super();
     this.box = true;
+    if (initValue instanceof DG.FuncCall) {
```
skalkin, line 35: Let's shorten it:

`this.funcName = initValue instanceof DG.FuncCall ? initValue.func.name : initValue;`

AufarZakiev: Even better, we can safely remove these lines. `funcName` is never used in the library and never used in third-paty code (just checked).

libraries/compute-utils/function-views/src/function-view.ts
```diff
@@ -192,13 +192,20 @@ export abstract class FunctionView extends DG.ViewBase {
    * @stability Stable
   */
   protected async loadFuncCallById() {
+    if (this.initValue instanceof DG.FuncCall) {
+      // next tick is needed to run funcCallReplaced before building UI
+      await new Promise(resolve => setTimeout(resolve, 0));
```
AufarZakiev, line 193: I can't get point of this line. 
`this.funcCallReplaced` only has meaning when UI is already built. Using this event, we can update UI when `this.funcCall` is replaced. By the moment of this line, UI is not built yet. 

Tried to comment this line and run my code - it works properly.

libraries/compute-utils/function-views/src/pipeline-view.ts
```diff
@@ -75,7 +75,14 @@ export class PipelineView extends ComputationView {
 
     this.subs.push(
       grok.functions.onAfterRunAction.pipe(
-        filter((run) => Object.keys(this.steps).includes(run.func.nqName)),
+        filter((run) => {
+          for (const {view} of Object.values(this.steps)) {
```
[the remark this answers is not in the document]
AufarZakiev: Could be replaced by 
```
filter((run) => Object.values(this.steps).some(({view}) => (view?.funcCall?.id === run?.id) && run)),
```

libraries/compute-utils/function-views/src/function-view.ts
AufarZakiev: Ah, got it

AufarZakiev (approved):

PR #1767, opened 2023-04-07
grok 12771 molfile parser

ghost:


libraries/chem-meta/src/parsing-utils/molfile-handler.ts
```diff
@@ -101,7 +110,25 @@ export class MolfileHandler extends ChemicalTableParserBase implements ChemicalT
...
+  private isQuote(idx: number): boolean {
+    return this.fileContent.at(idx) === '\"' || this.fileContent.at(idx) === '\'';
+  }
+
+  private getNextIdenticalChar(idx: number): number {
+    const sym = this.fileContent.at(idx);
+    if (sym)
+      return this.fileContent.indexOf(sym, idx + 1);
+    else
+      return -1;
```
LaykDimon, line 122: you can rewrite it as a ternary, it would be more brief and clear:
`sym ? return this.fileContent.indexOf(sym, idx + 1) : return -1;`

libraries/chem-meta/src/parsing-utils/chemical-table-parser.ts
```diff
@@ -1,5 +1,5 @@
 export interface ChemicalTableParser {
```
vdyma, line 1: A good convention is to name interfaces with prefix `I`: `IChemicalTableParser`

libraries/chem-meta/src/parsing-utils/chemical-table-parser-base.ts
```diff
@@ -19,26 +9,26 @@ type CoordinateArrays = {
...
 }
 
-/** Base singleton for Molfile or Mol2 parser/handler */
-export abstract class ChemicalTableParserBase implements ChemicalTableParser {
-  protected constructor(file: string) {
-    this.init(file);
+/** Base for Molfile or Mol2 parser/handler */
+export abstract class ChemicalTableParserBase {
+  constructor(fileContent: string) {
+    this.init(fileContent);
```
skalkin, line 15: What's the point of having this method instead of doing all initialization in the constructor?

libraries/chem-meta/src/parsing-utils/chemical-table-parser-base.ts
```diff
@@ -19,26 +9,26 @@ type CoordinateArrays = {
...
+  protected _bondTypes?: Uint16Array;
 
-  public init(file: string): void {
-    this.file = file.replace(/\r/g, '');
+  protected init(fileContent: string): void {
+    this.fileContent = fileContent.replace(/\r/g, '');
     this._atomCount = undefined;
     this._atomTypes = undefined;
     this._bondCount = undefined;
+    this._bondTypes = undefined;
```
skalkin, line 31: Why are we initializing it to undefined? I believe this is the default value. And in any case, if we really need it then it's better to combine declaration with initialization

libraries/chem-meta/src/parsing-utils/molfile-handler.ts
```diff
@@ -1,159 +1,17 @@
...
-    if (MolfileHandler.validateV3K(molfile))
-      return MOLFILE_VERSION.V3000;
-    else if (MolfileHandler.validateV2K(molfile))
-      return MOLFILE_VERSION.V2000;
+import {MolfileHandlerBase} from './molfile-handler-base';
+import {MolfileV2KHandler} from './molfile-v2k-handler';
+import {MolfileV3KHandler} from './molfile-v3k-handler';
+
+/** Defines the proper handler based on the molfile type  */
+export class MolfileHandler {
```
skalkin, line 6: Why have this class at all? Can't we rename MolHandlerBase to MolHandler, and add this static method there? One less class, a bit more clarity

libraries/chem-meta/src/parsing-utils/molfile-v2k-handler.ts
```diff
@@ -0,0 +1,65 @@
...
+import {MolfileHandlerBase} from './molfile-handler-base';
+import {V2K_CONST} from '../formats/molfile-v2k-const';
+
+export class MolfileV2KHandler extends MolfileHandlerBase {
+  constructor(molfile: string) {
+    super(molfile);
+  }
+
+  public static validate(molfile: string): boolean {
+    return (molfile.indexOf(V2K_CONST.HEADER) !== -1 &&
```
skalkin, line 11: Wrong indentation

PR #1768, opened 2023-04-10
Mdolotova/grok 12422/molecule fragments

MariaDolotova:


libraries/ml/src/viewers/activity-cliffs.ts
```diff
@@ -70,7 +69,7 @@ export async function getActivityCliffs(df: DG.DataFrame, seqCol: DG.Column, enc
   methodName: string, semType: string, tags: {[index: string]: string},
   seqSpaceFunc: (params: ISequenceSpaceParams) => Promise<ISequenceSpaceResult>,
   simMatrixFunc: (dim: number, seqCol: DG.Column, df: DG.DataFrame, colName: string,
-    simArr: DG.Column[]) => Promise<DG.Column[]>,
+    simArr: (DG.Column | null)[]) => Promise<(DG.Column | null)[]>,
```
vdyma, line 72: Here the `DG.Column` implicitly has type `any`: `DG.Column<any>`.
Does column type matter here and in the other places through the code or the column can contain any data?

packages/Chem/src/package.ts
```diff
@@ -603,7 +605,7 @@ export async function activityCliffs(df: DG.DataFrame, molecules: DG.Column, act
     return;
   }
   const axesNames = getEmbeddingColsNames(df);
-  if (df.rowCount > 500) {
+  if (df.rowCount > 10000) {
```
vdyma, line 608: This number could probably be saved in constant as well?

packages/Chem/src/utils/mol-creation_rdkit.ts
```diff
@@ -1,4 +1,7 @@
 import {RDModule, RDMol} from "@datagrok-libraries/chem-meta/src/rdkit-api";
+import { isMolBlock } from "./chem-common";
```
vdyma, line 2: Spaces inside the braces and using double quote for imports

PR #1777, opened 2023-04-11
Compute: Rich function view input customizations support

andrey-sharapov-sc:
- Adding beforeInputPropertyRender and afterInputPropertyRender
Subjects to allow mutation of properties and rendered controlls. For
example, hiding programmatically filled script inputs.

- Adding beforeRenderControlls Subject and exposing replaceControlls
and getRunButton. This will allow a full customization of controlls
for complex pipeline workflows.

- Input reactivity refactored into separate functions for better
readability/maintainability.


libraries/compute-utils/function-views/src/rich-function-view.ts
```diff
@@ -88,6 +102,24 @@ export class RichFunctionView extends FunctionView {
     return Promise.resolve();
   }
 
+  public beforeInputPropertyRender = new Subject<DG.Property>();
```
AufarZakiev, line 105: Aren't hooks better in this case?
I feel I miss some differences between using hooks (and overloading them) and using Subjects (and subscribing for them). 

The see differences in dealing with async functions. Are there any other differences?

AufarZakiev (approved):

libraries/compute-utils/function-views/src/rich-function-view.ts
AufarZakiev: However, we can 
1) create class derived from `RichFunctionView`
2) create an editor function for this derived class
3) pass this editor function to a computational script using `editor` tag

It could be very flexible this way, what do you think?

PR #1780, opened 2023-04-12
GROK-12824: Grok Connect: Added the ability to expand complex objects

PavloPolovyi:


connectors/serialization/src/main/java/serialization/ComplexTypeColumn.java
```diff
@@ -0,0 +1,256 @@
...
+        if (ofArray.isPrimitive()) {
+            List ar = new ArrayList();
+            int length = Array.getLength(array);
+            for (int i = 0; i < length; i++) {
+                ar.add(Array.get(array, i));
+            }
+            return ar.toArray();
+        }
+        else {
+            return (Object[]) array;
```
LaykDimon, line 221: you don't really need these extra braces after else, just write it as
```
else
  return (Object[]) array;
```

PavloPolovyi: It's incorrect style for java. https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used

PR #1828, opened 2023-04-20
GROK-12792: BiostructureViewer: Add tests for different formats

drizhina:


packages/BiostructureViewer/src/tests/molstar-preview-tests.ts
```diff
@@ -0,0 +1,51 @@
...
+  test,
+  expect
+} from "@datagrok-libraries/utils/src/test";
+import { _package, molecule3dNglView1 } from "../package";
+import { _packageName } from "./utils";
+
+const validFileNames = ['1bdq.pdb', '1bdq.sdf', 'dc.mol2',
+  '4tkx.mmcif', 'caffeine.xyz', 'grofile.gro', 'pdbqt.pdbqt'];
+
+category("MolstarPreview", () => {
```
LaykDimon, line 16: please follow one style of writing strings like `'this'`

drizhina: ah, something is wrong with my linter.

skalkin (commented):
We have many similar standard files in different formats under /Demo file share, can't we use them for testing instead of shipping more samples?

packages/BiostructureViewer/src/package.ts
```diff
@@ -133,29 +133,29 @@ export async function importPdb(fileContent: string): Promise<DG.DataFrame[]> {
...
+export function molecule3dNglView3(file: DG.FileInfo): DG.View {
   return previewBiostructure(file);
 }
 
 // eslint-disable-next-line max-len
 //tags: fileViewer, fileViewer-dsn6, fileViewer-brix, fileViewer-cube, fileViewer-cub, fileViewer-dx, fileViewer-dxbin, fileViewer-xplor, fileViewer-cns, fileViewer-mrc, fileViewer-map, fileViewer-ccp4
 //input: file file
 //output: view v
-export function molecule3dNglView4(file: any): DG.View {
+export function molecule3dNglView4(file: DG.FileInfo): DG.View {
```
skalkin, line 158: Why do we need four functions, why not just have one?

drizhina: honestly, I have no idea. it was like this. as I was told before some time different formats were opened by different viewers, and now all are combined. I would ask @tanas80

PR #1842, opened 2023-04-25
GROK-12818: Chem: Get fingerprints without saving mols

MariaDolotova:


packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
```diff
@@ -17,43 +18,91 @@ export class RdKitServiceWorkerSimilarity extends RdKitServiceWorkerBase {
...
-      case Fingerprint.Morgan:
-        for (let i = 0; i < this._rdKitMols.length; ++i) {
-          try{
-            if (!this._rdKitMols[i] || this._rdKitMols[i] && this._rdKitMols[i]?.is_qmol)
+          break;
+        case Fingerprint.Morgan:
+          for (let i = 0; i < this._rdKitMols.length; ++i) {
+            try {
+              if (!this._rdKitMols[i] || this._rdKitMols[i] && this._rdKitMols[i]?.is_qmol)
+                fps.push(null);
```
skalkin, line 38: Why would we push(null) instead of allocating an array of the right size upfront, and then just setting i-th element?

skalkin (commented):
Can't we make it simpler? I thought it would be just one additional parameter, but this is +100 loc

packages/Chem/src/package.ts
```diff
@@ -483,25 +486,47 @@ export async function chemSpaceTopMenu(table: DG.DataFrame, molecules: DG.Column
     return;
   }
 
-  const embedColsNames = getEmbeddingColsNames(table);
+  const allowedRowCount = methodName === 'UMAP' ? 100000 : 10000;
+  const fastRowCount = methodName === 'UMAP' ? 5000 : 2000;
```
LaykDimon, line 490: can we put `'UMAP'` to const? we use it twice or more

packages/Chem/src/package.ts
```diff
@@ -483,25 +486,47 @@ export async function chemSpaceTopMenu(table: DG.DataFrame, molecules: DG.Column
...
+      seqCol: molecules,
+      methodName: methodName,
+      similarityMetric: similarityMetric,
+      embedAxesNames: [embedColsNames[0], embedColsNames[1]],
+      options: options
+    };
+    const chemSpaceRes = await chemSpace(chemSpaceParams);
+    const embeddings = chemSpaceRes.coordinates;
+  
+    for (const col of embeddings) {
```
LaykDimon, line 510: no need in curly braces

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
```diff
@@ -17,43 +18,91 @@ export class RdKitServiceWorkerSimilarity extends RdKitServiceWorkerBase {
...
       switch (fingerprintType) {
-      case Fingerprint.Pattern:
-        for (let i = 0; i < this._rdKitMols.length; ++i) {
-          try {
-            if (!this._rdKitMols[i])
+        case Fingerprint.Pattern:
+          for (let i = 0; i < this._rdKitMols.length; ++i) {
+            try {
+              if (!this._rdKitMols[i])
+                fps.push(null);
```
LaykDimon, line 25: do we really need to push null?

packages/Chem/src/rdkit.worker.ts
```diff
@@ -36,5 +36,8 @@ ctx.addEventListener('message', async (e: any) => {
   } else if (op === WORKER_CALL.CONVERT_MOL_NOTATION) {
     const result = _rdKitServiceWorker!.convertMolNotation(args[0]);
     port.postMessage({op: op, retval: result});
+  }else if (op === WORKER_CALL.GET_FINGERPRINTS_MOLS_ON_FLY) {
```
LaykDimon, line 39: space after curly brace

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
MariaDolotova: Corrected

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
```diff
@@ -17,43 +18,91 @@ export class RdKitServiceWorkerSimilarity extends RdKitServiceWorkerBase {
...
       // nothing to do, bit is already 0
     }
     return fps!.map((el: any) => {
-      return {data: el, length: el ? el.length : 0};
+      return { data: el, length: el ? el.length : 0 };
+    });
+  }
+
+
+  getFingerprintsWithMolsOnFly(dict: string[], fingerprintType: Fingerprint) {
```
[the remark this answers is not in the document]
MariaDolotova: Function removed, logic moved to getFingerprints. Initially I thought that making two separate functions would be more readable than using additional ifs inside 'getFingerprints'. But indeed using one function is definitely better idea.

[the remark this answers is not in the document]
MariaDolotova: Function removed, logic moved to getFingerprints

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
```diff
@@ -17,43 +18,91 @@ export class RdKitServiceWorkerSimilarity extends RdKitServiceWorkerBase {
...
+    const fps: (Uint8Array | null)[] = [];
+    for (let i = 0; i < dict.length; ++i) {
+      const item = dict[i];
+      let mol;
+      if (!item || item === '')
+        fps.push(new Uint8Array());
```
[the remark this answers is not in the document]
MariaDolotova: Because for empty molecules we want to get an empty bitset insted of null fingerprint (null is used for malformed data)

packages/Chem/src/package.ts
MariaDolotova: Done

packages/Chem/src/package.ts
MariaDolotova: Done

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
MariaDolotova: This logic has been removed

packages/Chem/src/rdkit.worker.ts
MariaDolotova: Method was removed

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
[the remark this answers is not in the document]
skalkin: Why would we want an empty bitset? We want to exclude it from all further computations, there is literally no molecule, end of story :)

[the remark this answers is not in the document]
MariaDolotova: Since we add a column with fingerprints to an initial dataframe, number of rows in returned fingerprints column should match the initial dataframe. So we need to keep rows for empty/malformed molecules. After, of course, we do not include those malformed/empty data into analisys.
And since we want to separate empty molecules from malfomed molecules (to warn users only of malformed molecules indexes) we separate them - empty bitsets for empty, nulls - for malformed. 
Btw, not only undefined or '' matches empty molecules. We can also have empty molv2000 or molv3000 which is basically not an empty string. And in case of empty molfile rdkit itself creates empty fingerprint bitset.

packages/Chem/src/rdkit-service/rdkit-service.ts
```diff
@@ -84,17 +84,24 @@ export class RdKitService {
...
     const t = this;
-    return (await this._doParallel(
-      (i: number, _: number) => {
-        return t.parallelWorkers[i].getFingerprints(fingerprintType);
-      },
-      (data: any) => {
-        return [].concat(...data);
-      })).map(
+    const res = dict ?
+      await this._initParallelWorkers(dict, (i: number, segment: any) =>
```
skalkin, line 91: Why is segment "any"?

skalkin (commented):
Much better now - but please see my comments, and let's discuss it after you provide your comments for functions

packages/Chem/src/package.ts
```diff
@@ -410,7 +410,10 @@ export function diversitySearchTopMenu(): void {
 //tags: editor
 //input: funccall call
 export function SearchSubstructureEditor(call: DG.FuncCall) {
```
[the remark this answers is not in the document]
MariaDolotova: Done

packages/Chem/src/package.ts
```diff
@@ -410,7 +410,10 @@ export function diversitySearchTopMenu(): void {
...
-  
+  if (grok.shell.tv.dataFrame.rowCount > MAX_SUBSTRUCTURE_SEARCH_ROW_COUNT) {
+    grok.shell.warning(`Dataset is too large. Please use dataset with less than ${MAX_SUBSTRUCTURE_SEARCH_ROW_COUNT} rows`);
```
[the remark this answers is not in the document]
MariaDolotova: Done

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
```diff
@@ -10,50 +16,59 @@ export class RdKitServiceWorkerSimilarity extends RdKitServiceWorkerBase {
     super(module, webRoot);
   }
 
-  getFingerprints(fingerprintType: Fingerprint) {
-    if (this._rdKitMols === null)
+  getFingerprints(fingerprintType: Fingerprint, dict?: string[]): IFingerprint[] {
```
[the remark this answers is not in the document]
MariaDolotova: Added comments

packages/Chem/src/rdkit-service/rdkit-service.ts
MariaDolotova: Added types

packages/Chem/src/widgets/chem-substructure-filter.ts
```diff
@@ -174,6 +175,10 @@ export class SubstructureFilter extends DG.Filter {
    * that would simply apply the bitset synchronously.
    */
   async _onSketchChanged(): Promise<void> {
+    if (this.dataFrame!.rowCount > MAX_SUBSTRUCTURE_SEARCH_ROW_COUNT) {
```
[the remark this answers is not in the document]
MariaDolotova: Not sure how to better implement it. Think of disabling filter in attach() method. But seems that we cannot disable filters using API. Or am I mistaken?

packages/Chem/src/rdkit-service/rdkit-service.ts
```diff
@@ -84,17 +84,24 @@ export class RdKitService {
...
+        t.parallelWorkers[i].getFingerprints(fingerprintType, segment),
+        (data: any) => {
+          return [].concat(...data);
```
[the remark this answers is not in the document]
skalkin: Isn't "[].concat(...data)" simply equals to "data"?

js-api/src/chem.ts
```diff
@@ -582,6 +581,12 @@ export namespace chem {
       return imageHost;
     }
 
+    disableSketcher(message: string) {
```
skalkin, line 584: I don't like this. This is breaking the sketcher and there is no way to enable it. APIs should not be done like that. Also, a method like that should be documented. 

I suggest something like `setEnabled(error: string | null)`. And it shouldn't remove anything, just put a div on top (and then remove if needed)

js-api/src/chem.ts
```diff
@@ -158,11 +150,17 @@ export namespace chem {
     resized = false;
     _sketcherTypeChanged = false;
     _autoResized = true;
+    _type = currentSketcherType;
```
skalkin, line 153: How many fields and properties do we need? There is _type, sketcherType, currentSketcherType. Why so many?

packages/Chem/src/chem-searches.ts
```diff
@@ -85,9 +85,10 @@ function colInvalidated(col: DG.Column): Boolean {
     col.getTag(FING_COL_TAGS.invalidatedForVersion) == String(col.version));
 }
 
-async function _invalidate(molCol: DG.Column) {
+async function _invalidate(molCol: DG.Column, createMols: boolean) {
   if (!colInvalidated(molCol)) {
-    await (await getRdKitService()).initMoleculesStructures(molCol.toList());
+    if (createMols)
+      await (await getRdKitService()).initMoleculesStructures(molCol.toList());
     LastColumnInvalidated = molCol.name;
```
skalkin, line 92: LastColumnInvalidated - should be in camelCase

packages/Chem/src/chem-searches.ts
```diff
@@ -128,16 +129,17 @@ function saveFingerprintsToCol(col: DG.Column, fgs: Uint8Array[], fingerprintsTy
...
     if (fgsCheck)
       return fgsCheck;
     else {
-      await _invalidate(molCol);
-      const fingerprints = await (await getRdKitService()).getFingerprints(fingerprintsType);
+      await _invalidate(molCol, createMols);
+      const molecules = createMols ? undefined : molCol.toList();
+      const fingerprints = await (await getRdKitService()).getFingerprints(fingerprintsType, molecules);
       saveFingerprintsToCol(molCol, fingerprints, fingerprintsType);
       chemEndCriticalSection();
```
skalkin, line 144: Do we understand how it all works? What does "useSection" do? Is it really ok to call "chemEndCriticalSection" (whatever this is) if useSection == false?

packages/Chem/src/package.ts
```diff
@@ -407,8 +407,11 @@ export function diversitySearchTopMenu(): void {
 //name: SearchSubstructureEditor
 //tags: editor
 //input: funccall call
-export function SearchSubstructureEditor(call: DG.FuncCall) {
-  
+export function searchSubstructureEditor(call: DG.FuncCall) {
```
skalkin, line 410: Why do we need this function? What does it do, is anyone using it?

packages/Chem/src/rdkit-service/rdkit-service-worker-client.ts
```diff
@@ -22,8 +22,8 @@ export class RdKitServiceWorkerClient extends WorkerMessageBusClient {
   freeMoleculesStructures = async () =>
     this.call(WORKER_CALL.FREE_MOLECULES_STRUCTURES);
 
-  getFingerprints = async (fingerprintType: Fingerprint) =>
-    this.call(WORKER_CALL.GET_FINGERPRINTS, [fingerprintType]);
+  getFingerprints = async (fingerprintType: Fingerprint, dict?: string[]) =>
```
skalkin, line 25: Why is it called "dict" if it's not a dictionary?

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
```diff
@@ -1,6 +1,12 @@
-import {RdKitServiceWorkerBase} from './rdkit-service-worker-base';
-import {defaultMorganFpLength, defaultMorganFpRadius, Fingerprint} from '../utils/chem-common';
-import {RDModule} from '@datagrok-libraries/chem-meta/src/rdkit-api';
+import { RdKitServiceWorkerBase } from './rdkit-service-worker-base';
+import { defaultMorganFpLength, defaultMorganFpRadius, Fingerprint } from '../utils/chem-common';
+import { RDModule } from '@datagrok-libraries/chem-meta/src/rdkit-api';
+import { getMolSafe, IMolContext } from '../utils/mol-creation_rdkit';
+
+export interface IFingerprint {
```
skalkin, line 6: Can't we infer length from the array?

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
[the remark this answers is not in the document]
skalkin: Regarding "Calculates fingerprints either on pre-created array of RDMols or creating RDMOls on the fly."

How do I pass "pre-created array of RDMols"? I only see a string[] parameter.
Oh, I think I get it. But please update the documentation to let others know how to populate the "pre-created array of RDMols".

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
```diff
@@ -10,50 +16,64 @@ export class RdKitServiceWorkerSimilarity extends RdKitServiceWorkerBase {
...
+              if (rdMol)
+                fps[i] = rdMol.get_pattern_fp_as_uint8array();
+            } catch {
+              //do nothing, fp is already null
+            }
+            break;
+          case Fingerprint.Morgan:
+            try {
+              if (rdMol && !isQMol)
+                fps[i] = rdMol.get_morgan_fp_as_uint8array(JSON.stringify({
```
skalkin, line 55: JSON.stringify(...) should be created outside of the loop and reused. It's very minor, but good style nonetheless.

skalkin (approved):
Approved, but please address my feedback

skalkin (approved):
Approved, but please address my feedback

js-api/src/chem.ts
MariaDolotova: As we discussed yesterday, I will implement disabling outside chem.ts bot to break API

js-api/src/chem.ts
MariaDolotova: Since saving this _type is onle neede for substructure filters synchronization, I've removed both _type and sketcherType getter from Sketcher object. Save it on substructure filter side.

packages/Chem/src/chem-searches.ts
MariaDolotova: Done

packages/Chem/src/package.ts
MariaDolotova: This is used by SubstructureSearchTopMenu function. I used this custom editor because in case we have only one molecule column we do not want function editor to open. It should automatically grab this column and open sketcher. If I'm not mistaken, in case we use just function with input molecule column parameter the input dialog will always be opened before running the fucntion. If it is ok, I will remove function editor.

packages/Chem/src/rdkit-service/rdkit-service-worker-client.ts
MariaDolotova: Frankly, I just used the same name as it was in initMoleculeStructures(). Can't tell why it was named that way. But agree, it should be renamed. Will do.

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
MariaDolotova: Yes, we definitely can. I removed length both with IFingerprint interface. We do not need it anymore.

packages/Chem/src/rdkit-service/rdkit-service-worker-similarity.ts
MariaDolotova: Done

packages/Chem/src/chem-searches.ts
MariaDolotova: Yes, it is ok, because as I see chemEndCriticalSection (which on its side calls criticalSectionEnd() fucntion) performs additional check for existance of exact key in global dictionary. And in case we didn't enter chemBeginCriticalSection() before caling chemEndCriticalSection we will not fail. But anyway we need to explore this part of code more.

PR #1856, opened 2023-05-01
GROK-12817: Dendrogram: Add loader to dendrogram

drizhina:


packages/Dendrogram/src/apps/hierarchical-clustering-sequences-app.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (this.viewed) {
+      await this.destroyView();
+      this.viewed = false;
+    }
+
+    this._df = df;
+
+    if (!this.viewed) {
+      await this.buildView();
+      this.viewed = true;
```
LaykDimon, line 40: you don't really need this if, because you'll always get into it
if you have true from the beginning, you'll at first destroy the view and then build new one
if you have false from the beginning, then you'll build the view
so if here is unnecessary

packages/Dendrogram/src/utils/hierarchical-clustering.ts
```diff
@@ -2,47 +2,73 @@ import * as grok from 'datagrok-api/grok';
...
-    .addButton('Ok', async () => { await exec(dlg); })
-    .show();
+import {attachLoaderDivToGrid} from '.';
+
+// Custom UI Dialog for Hierarchical Clustering
+export async function hierarchicalClusteringDialog(): Promise<void> {
+  let currentTableView = grok.shell.tv.table;
+  let currentSelectedColNames: string[] = [];
+
+  const availableColNames = (table:DG.DataFrame): string[] => {
```
LaykDimon, line 17: no space after `:`

packages/Dendrogram/src/utils/hierarchical-clustering.ts
```diff
@@ -2,47 +2,73 @@ import * as grok from 'datagrok-api/grok';
...
+    return table.columns.toList()
+      .filter((col) => col.type === DG.TYPE.FLOAT || col.type === DG.TYPE.INT || col.semType === 'Macromolecule')
```
LaykDimon, line 19: change `'Macromolecule'` to const

packages/Dendrogram/src/viewers/tree-renderers/canvas-tree-renderer.ts
```diff
@@ -354,18 +357,7 @@ export class CanvasTreeRenderer<TNode extends MarkupNodeType>
...
-        this.clickCounter += 1;
-        if (this.clickCounter === 2) {
-          this.clickCounter = 0;
-          this.onResetZoom();
-        }
-        setTimeout(() => {
-          this.clickCounter = 0;
-        }, 300);
-      }
+      } else { this.current = this.mouseOver; }
```
LaykDimon, line 360: you don't need braces here

drizhina: We have the following eslint rule: which makes sure that braces are consistent in blocks. i.e. the above if statement of this 'else' is big so it uses braces and so should this else. I don't think its good idea to remove this rule, we have it in many places.
image

packages/Dendrogram/src/apps/hierarchical-clustering-sequences-app.ts
drizhina: true, that bit of code I copied over and didn't pay much attention... thanx!

packages/Dendrogram/src/viewers/tree-renderers/canvas-tree-renderer.ts
LaykDimon: hm. guess we need to discuss this true on code review
because in different places we have different rules, such as writing like `{func}` or like `{ func }`

PR #1862, opened 2023-05-03
oserhiienko: grok-12613: signature editor refactoring

Aleksashka11:


packages/DevTools/src/signature-editor/const.ts
```diff
@@ -0,0 +1,215 @@
...
+    case FUNC_PROPS_FIELD.DESCRIPTION: return 'Description';
+    case FUNC_PROPS_FIELD.LOGIN: return 'Author';
+    case FUNC_PROPS_FIELD.HELP_URL: return 'Help URL';
+    case FUNC_PROPS_FIELD.LANGUAGE: return 'Language';
+    case FUNC_PROPS_FIELD.REFERENCE: return 'Reference';
+    case FUNC_PROPS_FIELD.SAMPLE: return 'Sample';
+    case FUNC_PROPS_FIELD.ENVIRONMENT: return 'Environment';
+    case FUNC_PROPS_FIELD.TAGS: return 'Tags';
+  }
+};
```
LaykDimon, line 52: instead of using `switch` you can try to make something like this:
`
export const FUNCTION_PROPS_LABELS: {[key: FUNC_PROPS_FIELD]: string} = {
  FUNC_PROPS_FIELD.NAME: 'Name',
  FUNC_PROPS_FIELD.CONNECTION: 'Connection'
  ...
};
`
so then you just need to write `FUNCTION_PROPS_LABELS[someVariable]` and you'll get the value just by accessing it by property name

packages/DevTools/src/signature-editor/const.ts
```diff
@@ -0,0 +1,215 @@
...
+export const getChoicesByName = (name) => {
+    switch (name) {
+        case FUNC_PROPS_FIELD.LANGUAGE: return {choices: languages};
+        case FUNC_PROPS_FIELD.DESCRIPTION: return {editor: 'textarea'};
+        case FUNC_PARAM_FIELDS.DIRECTION: return {choices: [...Object.values(DIRECTION)]};
+        case FUNC_PARAM_FIELDS.TYPE: return {choices: funcParamTypes};
+        case OPTIONAL_TAG_NAME.COLUMNS: return {choices: ['numerical', 'categorical']};
+        case OPTIONAL_TAG_NAME.TYPE: return {choices: ['numerical', 'categorical', 'dateTime']}; 
+    }
+}
```
LaykDimon, line 64: so you can use the same approach here and in other places if you want

packages/DevTools/src/signature-editor/function-signature-editor.ts
```diff
@@ -3,285 +3,88 @@ import * as ui from 'datagrok-api/ui';
...
   }, 500);
 }
 
 function addFseRibbonScript(v: DG.View) {
   setTimeout(() => {
     const panels = v.getRibbonPanels();
-    // @ts-ignore
-    const iconFse = ui.iconFA('magic', () => openFse(v, v.root.lastChild.lastChild.CodeMirror.getDoc().getValue()), 'Open Signature Editor');
+    //@ts-ignore
+    const iconFse = ui.iconFA('magic', () => openFse(v, v.root.querySelector('.CodeMirror').CodeMirror.getDoc().getValue()), 'Open Signature Editor');
```
LaykDimon, line 35: you can use `getElementsByClassName('CodeMirror)` instead of `querySelector`

packages/DevTools/src/signature-editor/function-signature-editor.ts
```diff
@@ -3,285 +3,88 @@ import * as ui from 'datagrok-api/ui';
...
+  conn = connectionName;
   const query = connection.query('sql', sql);
   return query;
 }
 
+async function getInputCode(v: DG.View, functionCode: string) {
+  let inputCodeCopy;
+  v.type === DATA_QUERY_VIEW 
+  ? inputCodeCopy = await getDataQuery(functionCode) 
+  : inputCodeCopy = DG.Script.create(functionCode);
```
LaykDimon, line 72: you don't need to create null variable `inputCodeCopy`, you can assign it to a ternary:
```
const inputCodeCopy = v.type === DATA_QUERY_VIEW ?
  await getDataQuery(functionCode) :
  DG.Script.create(functionCode);
```

packages/DevTools/src/signature-editor/function-signature-editor.ts
```diff
@@ -651,18 +343,19 @@ async function openFse(v: DG.View, functionCode: string) {
...
+
     if (gc.isTableCell) {
-      gc.style.element =
-        ui.divH([
-          deleteBtn(gc.grid.dataFrame?.get('Name', gc.gridRow)),
-        ]);
+      gc.style.element = ui.divH([
+        deleteBtn(gc.gridRow),
+        addParamBtn(gc.cell.rowIndex)
+      ], {style: {'width': '100px'}});
```
LaykDimon, line 358: please consider moving styles to CSS files

packages/DevTools/src/signature-editor/function-signature-editor.ts
skalkin: also no need for //ts-ignore, just cast it to `any` since this is exactly what's happening

PR #1871, opened 2023-05-05
GROK-13027: Bio | Tools: Fix Molecule3D grid cell renderer

tanas80:
Fix Molecule3D grid cell renderer in preview.
An erroneous design docked NGL drawing service component along a current grid. This is available for DG.TableView only and is not available for preview view. Also, this leads to the recreation of the NGL instance for the next DG.TableView opened.
Now we use a single NglGlDocService at the document level.


packages/BiostructureViewer/src/package.ts
```diff
@@ -104,7 +105,7 @@ declare const window: BsvWindowType;
 //output: object result
 export function getNglGlService(): NglGlServiceBase {
   if (!(window.$nglGlService)) {
-    const svc: NglGlService = new NglGlService();
+    const svc: NglGlServiceBase = new NglGlDocService();
```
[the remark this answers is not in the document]
skalkin: I wonder what the name stands for?

 NglGlServiceBase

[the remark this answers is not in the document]
skalkin: Why can't it be a simple exported variable?

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
```diff
@@ -0,0 +1,210 @@
...
+import * as DG from 'datagrok-api/dg';
+
+import {_package} from '../package';
+import * as NGL from 'NGL';
+import {NglGlServiceBase, NglGlTask} from '@datagrok-libraries/bio/src/viewers/ngl-gl-viewer';
+
+
+export class NglGlDocService implements NglGlServiceBase {
+  private readonly nglDiv: HTMLDivElement;
+  private readonly ngl: any;
```
skalkin, line 12: Is it really "any"? Can we get types for NGL?

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
```diff
@@ -0,0 +1,210 @@
...
+  constructor() {
+    const r = window.devicePixelRatio;
+
+    this.nglDiv = ui.div([], 'd4-ngl-viewer');
+    this.ngl = new NGL.Stage(this.nglDiv);
+    this.ngl.viewer.signals.rendered.add(this.onNglRendered.bind(this));
+
+    // The single NGL component
+    this.hostDiv = ui.div([this.nglDiv]);
+    this.hostDiv.style.position = 'absolute';
```
skalkin, line 30: Consider using css styles. Consider hiding it by setting display to none.

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
```diff
@@ -0,0 +1,210 @@
...
+        // stalled task
+        this._queue.splice(qI);
+        delete this._queueDict[key!];
+        swept = true;
+      }
+    }
+
+    this._busy = this._queue.length > 0;
+    if (!this._busy && swept) {
+      // Some tasks had been swept
```
skalkin, line 122: And?

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
```diff
@@ -0,0 +1,210 @@
...
+
+      // TODO: Convert string to Blob once converting PDB string column to Blob
+      const stringBlob = new Blob([task.props.pdb], {type: 'text/plain'});
+
+      this.task = undefined;
+      this.key = undefined;
+      // this.nglDiv.style.width = `0px`;
+      // this.nglDiv.style.height = `0px`;
+      // await this.ngl.handleResize();
+      // TODO: Use canvas size switching 0/1 px to required
```
skalkin, line 80: Please remove unnecessary comments before submitting a PR. (if you really intend to use them later, ok to keep). But in general, in order to get quality feedback you should make reviewer's life as easy as possible

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
```diff
@@ -0,0 +1,210 @@
...
+  private async _processQueue() {
+    const {key, task} = this._queue.shift()!;
+    if (key) delete this._queueDict[key];
+    try {
+      const r = window.devicePixelRatio;
+
+      // TODO: Convert string to Blob once converting PDB string column to Blob
+      const stringBlob = new Blob([task.props.pdb], {type: 'text/plain'});
+
+      this.task = undefined;
```
skalkin, line 75: Why are you setting it to undefined if they are reset again on line 93?

packages/BiostructureViewer/src/utils/pdb-grid-cell-renderer.ts
```diff
@@ -120,15 +120,20 @@ export class PdbGridCellRenderer extends DG.GridCellRenderer {
...
-
-            const imageCache = gridCell.tableColumn!.temp.get(PDB_RENDERER_IMAGE_CACHE_KEY);
-            imageCache[rowIdx] = image;
-            gridCell.tableColumn!.temp.set(PDB_RENDERER_IMAGE_CACHE_KEY, imageCache);
-
-            service.renderOnGridCell(g, new DG.Rect(x, y, w, h), gridCell, image);
+            g.save();
+            try {
+              _package.logger.debug('PdbRenderer.render() onAfterRender() ' + `rowIdx = ${rowIdx}`);
+              const imageStr: string = canvas.toDataURL();
```
skalkin, line 126: This makes no sense to me. We have two canvases, we want to render the content of canvas 1 on top of canvas 2. Why are we creating a string for it, can't we do it more efficiently?

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
```diff
@@ -0,0 +1,210 @@
+import * as grok from 'datagrok-api/grok';
+import * as ui from 'datagrok-api/ui';
+import * as DG from 'datagrok-api/dg';
+
+import {_package} from '../package';
+import * as NGL from 'NGL';
+import {NglGlServiceBase, NglGlTask} from '@datagrok-libraries/bio/src/viewers/ngl-gl-viewer';
+
+
+export class NglGlDocService implements NglGlServiceBase {
```
skalkin, line 10: Why do we need a "service", why can't it be a simple exported function?

skalkin (changes requested):
Hi Alexander, when submitting a PR with the non-trivial code please describe the problem statement (what you are solving), and the approach chosen (how). Otherwise you are risking getting only a trivial feedback about formatting, etc :)

packages/BiostructureViewer/src/package.ts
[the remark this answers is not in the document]
tanas80: To be sure there is a single instance of these NGL drawing service in the document.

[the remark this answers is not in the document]
tanas80: > what the name stands for? NglGlServiceBase

'GlService' stands for we handle for WebGl service. 'Base' postfix stands for an abstract class with a part of the common implementation. Leading 'Ngl' stands for specific implementation with the NGL library, as I assume we will adopt molstar for that purpose in the future.

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
tanas80: I added some typization with NGL.Stage class at typings/index.d.ts module.

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
```diff
@@ -0,0 +1,210 @@
...
+  private readonly nglDiv: HTMLDivElement;
+  private readonly ngl: any;
+
+  private readonly hostDiv: HTMLDivElement | null = null;
```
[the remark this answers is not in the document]
tanas80: Fixed

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  private readonly _queue: { key?: keyof any, task: NglGlTask, dt: number }[];
```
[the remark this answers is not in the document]
tanas80: I conceived this service as a general purpose service. In this case, it draws molecules for grid rows. But generally, the service can be used for other tasks. By design, you can not specify the key for the task, then the service will not throw out outdated tasks for drawing with the same key.

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
tanas80: style.display='none' disables NGL drawing at all

css: There is a single use of these settings "hiding" the div. Do you insist on using css style?

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
tanas80: Fixed.

packages/BiostructureViewer/src/utils/ngl-gl-doc-service.ts
tanas80: I use this value as the flag in the event handler onNglRendered() to prevent redundant copying of a white space drawing to the grid canvas (it works unstable, though).

packages/BiostructureViewer/src/utils/pdb-grid-cell-renderer.ts
tanas80: This was for debugging purposes. I removed these debug remnants.

PR #1930, opened 2023-05-16
GROK-13101: Bio: Add alignment options to kalign

drizhina:


packages/Bio/src/utils/multiple-sequence-alignment-ui.ts
```diff
@@ -33,20 +35,34 @@ export async function multipleSequenceAlignmentUI(options: multipleSequenceAlgin
...
-    // UI
+    // UI for PepSea alignment
     const methodInput = ui.choiceInput('Method', options.pepsea.method, pepseaMethods);
     methodInput.setTooltip('Alignment method');
+
+    // UI for Kalign alignment
+    const terminalGapInput = ui.floatInput('Terminal gap', options?.kalign?.terminalGap ?? null);
+    terminalGapInput.setTooltip('Penalty for opening a gap at the beginning or end of the sequence');
+    const kalignVersionDiv = ui.p(`Kalign version: ${kalignVersion}`,
+      {style: {marginBottom: 0, textAlign: 'end', fontSize: '12px', opacity: 0.8}}
```
LaykDimon, line 46: better move the styles into the CSS file

packages/Bio/src/utils/multiple-sequence-alignment-ui.ts
```diff
@@ -161,3 +180,21 @@ async function onColInputChange(
...
 }
+
+type MSADialogType = 'kalign' | 'pepsea';
+
+function switchDialog(
+  pepseaInputRootStyles: CSSStyleDeclaration[], kalignInputRootStyles: CSSStyleDeclaration[], dialogType: MSADialogType
+) {
+  if (dialogType === 'kalign') {
+    for (const inputRootStyle of pepseaInputRootStyles)
+      inputRootStyle.display = 'none';
```
LaykDimon, line 191: I see you're trying to change the style of every root element
firstly, change not the display value but just assign or unassign the class name
also, I can suggest you a better solution
maybe instead of changing the display of every element you better change the display value of the whole container, where these roots are stored?
and in this container you will refer to every root element using CSS `>` selector?

PR #1955, opened 2023-05-23
GROK-13139: Chem: Move MCS calculation to webworker

drizhina:


packages/Chem/src/utils/most-common-subs.ts
```diff
@@ -1,53 +1,23 @@
...
-
-    for (let j = 0; j < mols.length; j++)
-      mols[j].delete();
-
-    return smarts;
-  }
-
-  return "";
+  return new Promise((resolve, reject) => {
+    const worker = new Worker(new URL('./most-common-subs-worker.ts', import.meta.url));
```
StLeonidas, line 13: It should be implemente in RDKit service

skalkin (commented):
Guys, not sure if we still need it or what the status is? @drizhina feel free to do what you want with it :)

PR #1958, opened 2023-05-23
GROK-13165: Math-Lib: Create math lib and add clustering

drizhina:
first commit for math library. in this PR, math library is created and hierarchical clustering script added to it.
after the lib is merged and published, next step is to remove the same wasm script from dendrogram package, and make use of the math library. following will be the extensions of this library with other high performance computing scripts


libraries/math/src/hierarchical-clustering/wasm/clustering-worker-creator.ts
```diff
@@ -0,0 +1,20 @@
+import {ClusterMatrix} from '../types';
+
+/** Calculates Cluster matrix based on distance matrix reduces array. i.e. the top right triangle of the matrix
+ * turned into the array.
+ * @param distMatArray - distance matrix reduces array.
+ * @param n - dimension of the distance matrix, i.e. side length.
+ * @param methodCode - method code for the clustering algorithm.
```
vmakarichev, line 7: I think that some notes about methodCode should be added.
For example, 0 - cool method, 1 - better method, 2 - the best method, etc.

libraries/math/src/hierarchical-clustering/wasm/cpp/clustering.cpp
```diff
@@ -0,0 +1,64 @@
...
+//     delete[] clusters;
+//     return out;
+// }
+
+EMSCRIPTEN_KEEPALIVE
+void getDendrogram(float *distmat, int npoints, int optMethod, int *merge, float *heights){
+
+    hclust_fast(npoints, distmat, optMethod, merge, heights);
+    // don't forget to delete merges and heights from the client;
+    free(distmat);
```
vmakarichev, line 62: Compare this line and the line 42 from the file libraries/math/src/hierarchical-clustering/wasm/clusterizerWasm.js
I think that there is a multiple usage of "free" for the same pointer.

libraries/math/src/hierarchical-clustering/wasm/cpp/emsc command.txt
```diff
@@ -0,0 +1 @@
+em++ -O3 ./clustering.cpp ./fastcluster.cpp -o ./wasmCluster.js -s TOTAL_MEMORY=268435456 -s WASM=1 -s ALLOW_MEMORY_GROWTH=1 -s MODULARIZE=1 -s EXPORT_NAME="exportCppLib" -s EXPORTED_FUNCTIONS=["_getDendrogram","_malloc","_free"] -s EXPORTED_RUNTIME_METHODS=["cwrap","ccall"] -sENVIRONMENT=web,worker
```
vmakarichev, line 1: GREAT THANKS for presenting the command applied!

libraries/math/src/hierarchical-clustering/wasm/cpp/fastcluster.cpp
```diff
@@ -0,0 +1,191 @@
+//
```
vmakarichev, line 1: I think that we should discuss with @skalkin the following:
are we allowed to present such files in our repository?

drizhina: we had a discussion actually when I was adding this in dendrogram package. we decided to preserve the code along with its license. as long as its not too large

libraries/math/README.md
```diff
@@ -0,0 +1,3 @@
+# math
+
+math is a [library](https://datagrok.ai/help/develop/develop#libraries) for the [Datagrok](https://datagrok.ai) platform.
```
StLeonidas, line 3: Please give more specific description

PR #1966, opened 2023-05-24
GROK-13104: Improve logging of queries

PavloPolovyi:


connectors/grok_connect/src/main/java/grok_connect/utils/QueryManager.java
```diff
@@ -100,6 +100,7 @@ public DataFrame getSubDF(int dfNumber) throws IOException, SQLException, QueryC
         if (!connection.isClosed() && !resultSet.isClosed()) {
             df = getResultSetSubDf(dfNumber, columns);
             if (isDryRun) {
+                serializeDf(df, dfNumber);
```
skalkin: Why are we "serializing" the df for the dry run?

connectors/grok_connect/src/main/java/grok_connect/utils/QueryManager.java
```diff
@@ -113,6 +114,7 @@ public DataFrame getSubDF(int dfNumber) throws IOException, SQLException, QueryC
                     column.empty();
                 }
                 df = getResultSetSubDf(dfNumber, columns);
+                serializeDf(df, dfNumber);
```
skalkin: Why are we ignoring the result? Is it all done for testing purposes only?

PR #1979, opened 2023-05-29
GROK-13222: Bio: Fix all lint errors and warnings

drizhina:


packages/Bio/src/tests/activity-cliffs-utils.ts
```diff
@@ -1,11 +1,13 @@
...
-import {delay, expect} from '@datagrok-libraries/utils/src/test';
+import {expect} from '@datagrok-libraries/utils/src/test';
 import {activityCliffs} from '../package';
-import { DimReductionMethods } from '@datagrok-libraries/ml/src/reduce-dimensionality';
+import {DimReductionMethods} from '@datagrok-libraries/ml/src/reduce-dimensionality';
 
-export async function _testActivityCliffsOpen(df: DG.DataFrame, numberCliffs: number, method: DimReductionMethods, colName: string) {
+export async function _testActivityCliffsOpen(
+  df: DG.DataFrame, numberCliffs: number, method: DimReductionMethods, colName: string
+) {
```
vdyma, line 10: Please, do not add unnecessary lines. This could have been written in two lines: put as many parameters as you can on the line with the function name and the rest on the next line, including the closing bracket.
Also if the line ends with a parameter, it should have a comma in the end, no matter if it's the last parameter.

packages/Bio/src/tests/detectors-benchmark-tests.ts
```diff
@@ -23,35 +21,35 @@ category('detectorsBenchmark', () => {
   // -- fasta --
 
   test('fastaDnaShorts50Few50', async () => {
-      const et: number = await detectMacromoleculeBenchmark(10, NOTATION.FASTA, ALPHABET.DNA, 50, 50);
-    },
-    {skipReason: '#1192'});
+    const _et: number = await detectMacromoleculeBenchmark(10, NOTATION.FASTA, ALPHABET.DNA, 50, 50);
```
vdyma, line 24: This variable seem to be never used and doesn't provide any info. I guess we can remove it.

packages/Bio/src/tests/mm-distance-tests.ts
```diff
@@ -120,7 +120,7 @@ category('Distance', async () => {
       const seq1 = Array(10000).fill('FWRY').join('');
       const seq2 = Array(10000).fill('FYWRRY').join('');
       _testDistance(seq1, seq2, df, -20000);
-    } else _testDistance(prot5, prot6, df, 1);
+    } else { _testDistance(prot5, prot6, df, 1); }
```
vdyma, line 123: Put expressions in if-else on the next line and don't use brackets if it's a single line.

packages/Bio/src/tests/sequence-space-test.ts
```diff
@@ -14,7 +14,9 @@ category('sequenceSpace', async () => {
   let testHelmWithEmptyRowsTableView: DG.TableView;
 
   test('sequenceSpaceOpens', async () => {
-    testFastaDf = await readDataframe(DG.Test.isInBenchmark ? 'test/peptides_motif-with-random_10000.csv':'tests/100_3_clustests.csv');
+    testFastaDf = await readDataframe(
+      DG.Test.isInBenchmark ? 'test/peptides_motif-with-random_10000.csv':'tests/100_3_clustests.csv'
```
vdyma, line 18: All operators should have spaces before and after (`:` in this case).
Also, this should be rewritten as follows:
```
testFastaDf = await readDataframe(DG.Test.isInBenchmark ? 'test/peptides_motif-with-random_10000.csv' :
  'tests/100_3_clustests.csv');
 ```
 Or alternatively, if it fits the line length limit:
 ```
testFastaDf = await readDataframe(
  DG.Test.isInBenchmark ? 'test/peptides_motif-with-random_10000.csv' : 'tests/100_3_clustests.csv');
 ```

File: packages/Bio/src/tests/utils.ts [hunk 5]
```diff
@@ -33,7 +32,9 @@ export function _testTableIsNotEmpty(table: DG.DataFrame): void {
   expect(table.columns.length > 0 && table.rowCount > 0, true);
 }
 
-/** Waits if container is not started */
+/** Waits if container is not started
+ * @param {number} ms - time to wait in milliseconds
+*/
```
vdyma, line 37: Close the block comment on the same line as the last parameter

packages/Bio/src/tests/mm-distance-tests.ts
drizhina: generally yes. but this is according to what is found in packages eslint config, which dictates that braces should be consistent. i.e. if the if statement on line 119 has braces, so should the consequent elses. this rule was there before me.

packages/Bio/src/tests/sequence-space-test.ts
drizhina: Done, also Added rule for such cases.

packages/Bio/src/tests/activity-cliffs-utils.ts
skalkin: > Also if the line ends with a parameter, it should have a comma in the end, no matter if it's the last parameter.

I don't think we should require a comma after the last parameter, why would we?

ddrjn: there is actually a good reason to this. for example, It is particular useful when adding, removing, or reordering items in a list that spans multiple lines, because it reduces the number of lines that need to be changed, which helps with both editing and reviewing the diff. you can also refer to this [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas)

skalkin: I am against forcing developers to use trailing commas "just in case". Adding a new one is easy, and the diff looks just fine. You link just says that it's "particularly useful when reodering" - well, sort of, but how often do you reorder? 

There are places when it's useful and I'm not against it. But this is totally not an issue and should not be a rule.

PR #1989, opened 2023-05-31
GROK-13077: No action on merge-commit: add check merge commit jobs to all workflows on push

ksumag:
Every workflow with trigger 'on push' was updated with trigger 'on pull request'  with the relevant path. Jobs, with check merge commit present, have been added.  Now the workflow check merge commit is present: the next jobs will not start if it has.


.github/workflows/docusaurus.yaml
```diff
@@ -8,11 +8,43 @@ on:
...
       - 'docusaurus/**'
       - 'js-api/**'
+  pull_request:
+    paths:
+      - 'help/**'
+      - 'docusaurus/**'
+      - 'js-api/**'
 
 jobs:
+  process-merge-commit:
```
dnillovna, line 18: As it is a common part for ALL workflows lets make it a reusable workflow instead https://docs.github.com/en/actions/using-workflows/reusing-workflows to avoid copy-paste

help/docusaurus.yaml
```diff
@@ -0,0 +1,217 @@
+name: Docusaurus
```
dnillovna, line 1: Is it a leftover? check your commits before pushing.

dnillovna (changes requested):

.github/workflows/docusaurus.yaml
```diff
@@ -8,11 +8,21 @@ on:
...
```
dnillovna, line 18: I would call it check-commit, process is not quite the right word here

.github/workflows/check_merge_commit.yaml
```diff
@@ -0,0 +1,27 @@
+name: Check merge commit
```
dnillovna, line 1: Let's create a separate subdirectory for such workflows which we include

PR #1990, opened 2023-05-31
Update MinimalLib with a version that exposes get_mcs() through MolIterator

ptosco:
@StLeonidas @MariaDolotova Here is a PR that updates `MinimalLib` with a version that exposes `get_mcs()` through `MolIterator` and returns a `JSMol` rather than a SMARTS string; this is preferable as not all query features can be exactly translated to SMARTS. One can always call `get_smarts()` on the returned `JSMol` if a SMARTS string is needed.
It also exposes additional `SubstructLibrary` functionality that will be useful for the migration that we discussed.

It would be great if you could confirm that this works for you and passes tests at your earliest convenience. Thanks in advance!


MariaDolotova:
Hi @ptosco ! Unfortunately we had to revert this PR.
During PR testing the follwoing issues have been found with get_mcs function:

1) It seems that function returns incorrect MCS. For example if you run the follwoing code snippet:

```
const csv = `smiles
  O=C1CN=C(c2ccccc2N1)C3CCCCC3
  CN1C(=O)CN=C(c2ccccc12)C3CCCCC3
  CCCCN1C(=O)CN=C(c2ccccc12)C3CCCCC3
  CC(C)CCN1C(=O)CN=C(c2ccccc12)C3CCCCC3
  O=C1CN=C(c2ccccc2N1CC3CCCCC3)C4CCCCC4
  CN1C(=O)CN=C(c2cc(Cl)ccc12)C3CCCCC3`

  const t = DG.DataFrame.fromCsv(csv);

  const res = await getMCS(t.col('smiles')!,true ,true); 
```
you will get the follwing molblock as a result:
image
We get the similar results (with all 'R's instead of atoms and zero bonds) for other example datasets as well. Also I tried to change parameters responsible for exact atoms/bonds serach but it seems that changes did not affect the result.

2) creating MolIterator requires new() operator here: 
image

In case it is not used we get the error: "BindingError: Use 'new' to construct MolIterator". If I add new operator then typescript gives an error "'new' expression, whose target lacks a construct signature".
@ts-ignore allows to build and run the code and it seems to create MolIterator correctly. But I believe that it should be fixed on minimal lib side.

ptosco:
@MariaDolotova Ah, no, I was wrong - I can reproduce the error at runtime. OK, will fix it.

PR #1993, opened 2023-06-01
Replace aromatic-utils and other refactoring

ptosco:
@StLeonidas @MariaDolotova
This is another intermediated cleanup/refactoring PR before switching to using `JSSubstructLibrary`.
This addresses the issue that @StLeonidas described in a PDF document he sent me a few days ago, and does some other cleanup:
- updated the TS definitions for MinimalLib
- updated MinimalLib
- removed all occurrences of `JSMol::is_valid()` as it is now unnecessary and deprecated
- replaced `aromatic-utils` with `JSMol::set_aromatic()` as it is simpler and faster
- refactored and simplified some logic

Please confirm that nothing breaks with this PR and I will then go ahead with implementing FP-assisted searches through `JSSubstructLibrary`.


ptosco:
@MariaDolotova Thanks for merging #1997. I have solved the conflict here; could you please check if this one works as expected? Many thanks!

PR #1997, opened 2023-06-01
Update MinimalLib with a version that exposes get_mcs() through MolIterator and returns a JSMol as MCS

ptosco:
- update MinimalLib with a version that exposes `get_mcs()` through `MolIterator` and returns a `JSMol` as MCS
- update RDKit API TS definitions


ptosco:
@MariaDolotova This PR fixes the issue with TS definitions.
Regarding the issue with molblock, that's expected. Molblock are not a good format for query atoms/bonds - many RDKit queries do not exactly translate to MDL queries.
A better format to serialize RDKit query is SMARTS (through `RDMol.get_smarts()`), but also that is not guaranteed to exactly recover all query features in the original molecule. If you need to _exactly_ serialize an RDKit molecule to recover it later, use `RDMol.get_as_uint8array()` (to serialize) and then `RDMol.get_mol_from_uint8array()` (to deserialize).

MariaDolotova:
Hi @ptosco ! Can you please check the issue with AtomCompare and BondCompare parameters. Changing the parameters doesn't affect the final result. For example if you run the following code you will get the same mcs for all cases - [#8]=[#6]1-[#6]-[#7]=[#6](-[#6]2:[#6](-[#7]-1):[#6]:[#6]:[#6]:[#6]:2)-[#6]1-[#6]-[#6]-[#6]-[#6]-[#6]-1

`const csv = `smiles
  O=C1CN=C(c2ccccc2N1)C3CCCCC3
  CN1C(=O)CN=C(c2ccccc12)C3CCCCC3
  CCCCN1C(=O)CN=C(c2ccccc12)C3CCCCC3
  CC(C)CCN1C(=O)CN=C(c2ccccc12)C3CCCCC3
  O=C1CN=C(c2ccccc2N1CC3CCCCC3)C4CCCCC4
  CN1C(=O)CN=C(c2cc(Cl)ccc12)C3CCCCC3`

  const t = DG.DataFrame.fromCsv(csv);

  console.log((await getMCS(t.col('smiles')!, true ,true))?.get_smarts()); 
  console.log((await getMCS(t.col('smiles')!, false, false))?.get_smarts());
  console.log((await getMCS(t.col('smiles')!, true, false))?.get_smarts());
  console.log((await getMCS(t.col('smiles')!, false, true))?.get_smarts());`

It seems that get_mcs always uses the default values for atoms and bond search types no matter which parameters are passed. I tried to look through C++ implementation and couldn't find what could go wrong there. Looks like everything should work as expexted but the results prove otherwise.

And one more question - how do you think, maybe it makes sense to return smarts string instead of RDMol object from get_mcs? If I understood correctly from your previous comment, neither get_molblock() nor get_smiles() are not expected to return correct results. But in case we return RDMol object we need to additionaly handle it in typescript code (I mean delete). Or are there reasons why it is better to return RDMol instead of just smarts string?

@StLeonidas , @skalkin also FYI.

PR #1999, opened 2023-06-02
GROK-13240: ML: Parallel Distance matrix calculation

drizhina:
Things done in this PR:

- Introduce DistanceMatrixService, which allows to calculate distance matrix on multiple workers. speedups up to 10x depending on CPU. 
- Use Condensed distance matrix(Float32Array) for dimensionality reduction methods instead of full nXn matrix.
- Fix tests
Results:
Dendrogram: clustering calculation speedup up to 10X on macromolecules.
Bio: Sequence space and activity cliffs calculation speedup ~5X. sim/div search speedup ~5-6X
Chem: unchanged. for now can not paralelize matrix calculation for BitArray, as webworkers can't be passed custom classes(coming in future)

### DIFF
--- a/libraries/ml/package.json
+++ b/libraries/ml/package.json
@@ -8,7 +8,7 @@
     "email": "vdyma@datagrok.ai"
   },
   "friendlyName": "Datagrok ML library",
-  "version": "6.3.28",
+  "version": "6.3.29",
   "description": "Machine learning supporting utilities",
   "dependencies": {
     "@datagrok-libraries/utils": "^4.0.8",
--- a/packages/Bio/package.json
+++ b/packages/Bio/package.json
@@ -5,7 +5,7 @@
     "name": "Leonid Stolbov",
     "email": "lstolbov@datagrok.ai"
   },
-  "version": "2.4.33",
+  "version": "2.4.34",
   "description": "Bioinformatics support (import/export of sequences, conversion, visualization, analysis). [See more](https://github.com/datagrok-ai/public/blob/master/packages/Bio/README.md) for details.",
   "repository": {
     "type": "git",
@@ -16,7 +16,7 @@
     "@biowasm/aioli": "^3.1.0",
     "@datagrok-libraries/bio": "^5.30.3",
     "@datagrok-libraries/chem-meta": "^1.0.1",
-    "@datagrok-libraries/ml": "^6.3.28",
+    "@datagrok-libraries/ml": "^6.3.29",
     "@datagrok-libraries/tutorials": "^1.3.2",
     "@datagrok-libraries/utils": "^4.0.8",
     "cash-dom": "^8.0.0",
--- a/packages/Chem/package.json
+++ b/packages/Chem/package.json
@@ -2,7 +2,7 @@
   "name": "@datagrok/chem",
   "friendlyName": "Chem",
   "fullName": "Chem",
-  "version": "1.6.15",
+  "version": "1.6.16",
   "author": {
     "name": "Leonid Stolbov",
     "email": "lstolbov@datagrok.ai"
@@ -44,7 +44,7 @@
   "dependencies": {
     "datagrok-api": "^1.14.0",
     "@datagrok-libraries/chem-meta": "^1.1.1",
-    "@datagrok-libraries/ml": "^6.3.25",
+    "@datagrok-libraries/ml": "^6.3.29",
     "@datagrok-libraries/utils": "^4.0.7",
     "@datagrok-libraries/tutorials": "^1.3.2",
     "cash-dom": "latest",


libraries/ml/src/distance-matrix/distance-matrix-service.ts
```diff
@@ -0,0 +1,70 @@
...
+          let lmin = 0;
+          let lmax = Number.MIN_VALUE;
+          for (let i = 0; i < this._workerCount; i++) {
+            const start = Math.floor(i * chunkSize);
+            const end = (i === this._workerCount - 1) ? totalLength : Math.floor((i + 1) * chunkSize);
+            const startRow = endRow;
+            const startCol = endCol;
+            if (i !== this._workerCount - 1) {
+              // These formulas map the linear index to the upper triangular matrix indices
+              endRow = len - 2 - Math.floor(Math.sqrt(-8 * end + 4 * len * (len - 1) - 7) / 2 - 0.5);
```
LaykDimon, line 37: guess you better put the numbers into constants, as the time will pass and you will forget what these numbers mean
and if you put it into named constants, you'll (and not only you) will recognize it much more better

libraries/ml/src/distance-matrix/distance-matrix.ts
```diff
@@ -58,8 +58,11 @@ export class DistanceMatrix {
     const size: number = list.length;
     const res = new DistanceMatrix(undefined, size);
     for (let i = 0; i < size; i++) {
-      for (let j = i + 1; j < size; j++)
-        res.set(i, j, method(list[i], list[j]));
+      for (let j = i + 1; j < size; j++) {
+        // if any of the values is null, set distance to 1
+        list[i] !== null && list[j] !== null ?
+          res.set(i, j, method(list[i], list[j])) : res.set(i, j, 1);
```
LaykDimon, line 64: you can do it even better
make the condition inside your set() method - `res.set(i, j, ..... ? method(list[I], list[j]) : 1)`

libraries/ml/src/reduce-dimensionality.ts
```diff
@@ -351,20 +382,20 @@ export class DimensionalityReducer {
...
    * @param {boolean} transpose Whether to transform coordinates to have columns-first orientation.
+   * @param {boolean} parallelDistanceWorkers Whether to use parallel distance computation.
    * @throws {Error} If the embedding method was not found.
    * @return {any} Cartesian coordinate of this embedding and distance matrix where applicable.
    * @memberof DimensionalityReducer
    */
-  public transform(transpose: boolean = false): { [key: string]: Matrix } {
-    if (this.reducer == undefined) {
+  public async transform(transpose: boolean = false, parallelDistanceWorkers?: boolean): Promise<IReduceDimensionalityResult> {
+    if (this.reducer == undefined)
```
LaykDimon, line 391: strict comparison

libraries/ml/src/spe.ts
```diff
@@ -84,9 +89,9 @@ export class SPEBase {
 
     let lambda2 = this.lambda2;
 
-    if (this.steps == 0) {
-      this.steps = vectors.length-1;
-    }
+    if (this.steps == 0)
```
LaykDimon, line 92: strict comparison

libraries/ml/src/spe.ts
```diff
@@ -213,19 +216,18 @@ export class OriginalSPE extends SPEBase {
 
     this.initDistance(vectors);
 
-    if (this.maxDistanceSteps == null) {
+    if (this.maxDistanceSteps == null)
```
LaykDimon, line 219: same

libraries/ml/src/viewers/activity-cliffs.ts
```diff
@@ -90,13 +90,13 @@ export async function getActivityCliffs(df: DG.DataFrame, seqCol: DG.Column, enc
...
   linesGridFunc?: (df: DG.DataFrame, pairColNames: string[]) => DG.Grid,
   seqSpaceOptions?: any, cliffsDockRatio?: number) : Promise<DG.Viewer> {
-
   activityCliffsIdx++;
   const similarityLimit = similarity / 100;
   const dimensionalityReduceCol = encodedCol ?? seqCol;
   let zoom = false;
   let ignoreSelectionChange = false;
   const cashedLinesData: any = {};
+  // eslint-disable-next-line prefer-const
```
LaykDimon, line 99: why is it done?

libraries/ml/src/distance-matrix/distance-matrix-service.ts
drizhina: I don't think that these numbers have any actual meaning. they are just some constants that pop out when you start working out how to map a linear indexed compressed distance matrix to a square one.

libraries/ml/src/reduce-dimensionality.ts
```diff
@@ -80,14 +85,15 @@ abstract class Reducer {
...
+  abstract transform(parallelDistanceWorkers?: boolean): Promise<IReduceDimensionalityResult>;
 }
 
 /** t-SNE dimensionality reduction. */
 class TSNEReducer extends Reducer {
   protected reducer: TSNE;
   protected iterations: number;
-  protected distanceFn: DistanceMetric;
+  protected distanceFname: KnownMetrics;
+  protected distanceFn: (a: any, b: any) => number;
```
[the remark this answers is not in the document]
drizhina: yes. distance function will take two arbitrary things and compute a distance. as far as reducer is concerned, they can be anything, strings, arrays, bitarrays, numbers...

libraries/ml/src/distance-matrix/distance-matrix-service.ts
LaykDimon: if we can't find their real meaning it's ok so

libraries/ml/src/reduce-dimensionality.ts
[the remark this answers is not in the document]
skalkin: How about 

```
protected distanceFn<T>: (a: T, b: T) => number
```
?

libraries/ml/src/spe.ts
skalkin: What's wrong with it though? We know that this.steps is a number, and we'd get a warning if we try to compare incompatible types using "=="

skalkin (approved):

libraries/ml/src/spe.ts
LaykDimon: why do we even need to use '==' in JS?
strict comparison is safer and faster (doesn't use type casting)

PR #2000, opened 2023-06-02
ComputeUtils: Custom object input render

andrey-sharapov-sc:


AufarZakiev (approved):
Elegant trick on bitwise compare

PR #2004, opened 2023-06-05
GROK-13229: SA: Variance-based sensitivity analysis

vmakarichev:
Sensitivity Analysis computations are implemented. Further, their megration to compute-utils is planned.


packages/SA/src/inputTools.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  for (let i = 0; i < length; ++i)
+    columnSource[i] = numInputInfo.min + step * randomData[i];
+
+  numInputInfo.column = column;
+  
+  return column;
+}
+
+export function createVariedNumericalInputColumns(samplesCount: number,
```
AufarZakiev, line 41: It is hard to debug functions when they vary input params. Is it possible to rewrite this function without changing inputs params with no performance penalty?

packages/SA/src/inputTools.ts
```diff
@@ -0,0 +1,50 @@
...
+
+import {getGeneratedColumnsData} from './inputsGeneration';
+
+export type VariedNumericalInputInfo = {
+  name: string,
+  caption: string | undefined,
+  type: DG.COLUMN_TYPE,
+  min: number,
+  max: number,
+  column: DG.Column | undefined
```
AufarZakiev, line 15: This structure could be simpler. 
`name`, `caption`, `type` could be stored directly in DG.Property

`column` field is always undefined before values generation. It is easier to create separate structure for filled info.
For example,
```
type VariedNumericalInputInfo = {
  prop: DG.Property,
  min: number,
  max: number,
}
```
and 
```
type VariedNumericalInputValues = VariedNumericalInputInfo & {column: DG.Column}
```

What do you think about it?

packages/SA/src/outputTools.ts
```diff
@@ -0,0 +1,99 @@
...
+
+type OutputInfo = {
+  name: string,  
+  type: DG.TYPE
+};
+
+function getOutputsInfo(func: DG.Func): OutputInfo[] {
+  const ouputInfo = Array<OutputInfo>(0);
+
+  for (const item of func.outputs)
```
AufarZakiev, line 17: Could be replaced in one-line by `map` function

AufarZakiev (changes requested):

packages/SA/src/sensitivityAnalysis.ts
```diff
@@ -0,0 +1,70 @@
...
+    for (const funcCall of this.funcCalls)
+      await funcCall.call();
+  }
+
+  async perform(): Promise<DG.DataFrame> {
+    await this.run();
+
+    const columns: DG.Column[] = [];
+
+    for (const input of this.variedInputs)
```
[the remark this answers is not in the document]
vmakarichev: Done

packages/SA/src/inputTools.ts
vmakarichev: Done

packages/SA/src/outputTools.ts
vmakarichev: Done

packages/SA/src/inputTools.ts
vmakarichev: Done

packages/SA/src/outputTools.ts
vmakarichev: GREAT THANKS FOR CODE REVIEW!

AufarZakiev (commented):
Will review later, after UI and computation polishing

PR #2008, opened 2023-06-05
GROK-13150: EDA | ANOVA: Integration

AnnaMuza:
One-way ANOVA computations implemented and integrated to EDA package


skalkin:
Please provide a description of what exactly is the intended effect, etc

packages/EDA/src/EDAtools.ts
```diff
@@ -44,3 +44,19 @@ export async function computePLS(table: DG.DataFrame, features: DG.ColumnList, p
...
   return _output;
 } 
+
+// Analysis of Variance (ANOVA)
+export async function computeANOVA(table: DG.DataFrame, columns: DG.ColumnList): Promise<number> 
+{
+  checkColumns(columns);
+
+  let _output: any;
+  let _promise = _oneWayAnovaInWebWorker(table, columns);
```
skalkin, line 54: Why would we do ANOVA in a web worker?

skalkin: I mean, it should be blazingly fast

packages/EDA/src/EDAtools.ts
```diff
@@ -44,3 +44,19 @@ export async function computePLS(table: DG.DataFrame, features: DG.ColumnList, p
 
   return _output;
 } 
+
+// Analysis of Variance (ANOVA)
+export async function computeANOVA(table: DG.DataFrame, columns: DG.ColumnList): Promise<number> 
+{
+  checkColumns(columns);
+
+  let _output: any;
```
skalkin, line 53: Specify the actual type instead of "any".

Name variable in camelCase, why the underscore?

packages/EDA/src/package.ts
```diff
@@ -55,6 +55,16 @@ export async function PLS(table: DG.DataFrame, names: DG.Column, features: DG.Co
   addPLSvisualization(table, names, features, predict, plsResults);
 }
 
+//top-menu: Tools | Data Science | Analysis of Variance (ANOVA)...
+//name: ANOVA
+//description: One-way Analysis of Variance (ANOVA)
+//input: dataframe table
+//input: column_list columns
+//output: double ftest 
+export async function ANOVA(table: DG.DataFrame, columns: DG.ColumnList): Promise<number> {
```
skalkin, line 64: What do "columns" represent?

packages/EDA/src/utils.ts
```diff
@@ -49,3 +49,13 @@ export function checkGeneratorSVMinputs(samplesCount: number, featuresCount: num
...
     throw new Error(INCORERRECT_PERCENTAGE_MES);
 }
+
+// Check columns (ANOVA)
+export function checkColumns(columns: DG.ColumnList): void {
+  if (Array.from(columns.numerical).length < columns.length)
+    throw new Error('non numerical columns are selected');
+
+  if (columns.toList().some((col) => col.stats.missingValueCount > 0))
+    grok.shell.warning('Columns with missing values are selected, the calculation results may not be accurate.\
```
skalkin, line 59: What is meant by "may not be accurate"? This is not right

packages/EDA/src/utils.ts
```diff
@@ -49,3 +49,13 @@ export function checkGeneratorSVMinputs(samplesCount: number, featuresCount: num
   if ((violatorsPercentage < PERCENTAGE_MIN) || (violatorsPercentage > PERCENTAGE_MAX))
     throw new Error(INCORERRECT_PERCENTAGE_MES);
 }
+
+// Check columns (ANOVA)
+export function checkColumns(columns: DG.ColumnList): void {
+  if (Array.from(columns.numerical).length < columns.length)
+    throw new Error('non numerical columns are selected');
```
skalkin, line 56: Start sentences with a capital case

packages/EDA/src/utils.ts
```diff
@@ -49,3 +49,13 @@ export function checkGeneratorSVMinputs(samplesCount: number, featuresCount: num
   if ((violatorsPercentage < PERCENTAGE_MIN) || (violatorsPercentage > PERCENTAGE_MAX))
     throw new Error(INCORERRECT_PERCENTAGE_MES);
 }
+
+// Check columns (ANOVA)
+export function checkColumns(columns: DG.ColumnList): void {
+  if (Array.from(columns.numerical).length < columns.length)
```
skalkin, line 55: Creating an array just to check whether there are non-numerical columns is expensive, rewrite it in a more efficient manner

skalkin (changes requested):

packages/EDA/wasm/ANOVA/ANOVA.cpp
```diff
@@ -0,0 +1,55 @@
+#include <iostream>
+#include "../../../../../Eigen/Eigen/Dense"
+#include <emscripten.h>
+
```
vmakarichev, line 4: I think that the following should be added:
1) problem specification, for example, "_... computation of smth the great importance is implemented .._."
2) links to scientific sources with formulas.

packages/EDA/src/EDAtools.ts
vmakarichev: Almost all parts of this code are generated automatically by C/C++-to-TS export script.
Now, all changes should be introduced manually. But further, I'll fix them in the export script.

vmakarichev: In other words, this is my fault.

packages/EDA/src/utils.ts
AnnaMuza: When passing columns with empty values to wasm functions, the empty values are converted to very small values close to zero (e.g., 2.78^-30), which is not NULL and can affect the calculations.

skalkin: Clearly they should not be converted to small values

packages/EDA/src/utils.ts
skalkin: ?

PR #2023, opened 2023-06-12
GROK-13302: JS API: Exposed popupMenu param for grok.chem.drawMolecule

vdyma:


skalkin (commented):
Not sure if we still need it?

vdyma:
> Not sure if we still need it?

That's not mandatory, but would be nice to have.

skalkin (approved):

PR #2047, opened 2023-06-20
Make the invalidated column key unique on table name

ptosco:
@StLeonidas @MariaDolotova This PR fixes a major bug which causes exceptions and misbehavior of the structure filter whenever there are multiple tables featuring molecule columns with the same name (e.g., `Structure`).
In such cases `lastColumnInvalidated` does not work as it should, and stale cache actually referring to a different column (but with the same name) is not invalidated, thus causing multiple problems (bogus results returned from structure filter) and exceptions thrown such as:
image

The fix is very simple: make the cache key unique on dataframe name _and_ column name.

### DIFF
--- a/packages/Chem/src/chem-searches.ts
+++ b/packages/Chem/src/chem-searches.ts
@@ -81,8 +81,12 @@ function _chemGetDiversities(limit: number, molStringsColumn: DG.Column, fingerp
   return diversities;
 }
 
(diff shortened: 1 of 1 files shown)


MariaDolotova:
Hi @ptosco ! Thank you for noticing and fixing. Have merged your PR.

PR #2062, opened 2023-06-22
Documentation: Change Scripts: On scripts dir and in files

ksumag:
Replace the directory 'Scripts' on 'scripts', in all files do so, then manually return back in it is not the path. 
I'm not sure about files in  selenium dir


dnillovna (changes requested):
Remove unnecessary changes

PR #2112, opened 2023-07-04
Mdolotova/grok 13172/substr search improvements

MariaDolotova:


packages/Chem/src/chem-searches.ts
```diff
@@ -82,116 +86,151 @@ function _chemGetDiversities(limit: number, molStringsColumn: DG.Column, fingerp
...
+function saveColumns(col: DG.Column, data: ((Uint8Array | null)[] | (string | null)[])[],
+  tags: string[], colTypes: DG.COLUMN_TYPE[], createdMols: boolean): void {
+  
+  for (let i = 0; i < data.length; i++)
+    saveColumn(col, data[i], tags[i], colTypes[i]);
+
+  const colVersion = createdMols ? col.version + data.length + 2 : col.version + data.length + 1;
+  if (createdMols)
+    col.setTag(FING_COL_TAGS.molsCreatedForVersion, String(colVersion));
+  for (let i = 0; i < data.length; i++)
```
skalkin, line 137: Not sure I follow the logic. Why is the for loop goes to data.length, but we are settings tags[i]? Shouldn't it go to tags.length?

packages/Chem/src/chem-searches.ts
```diff
@@ -82,116 +86,151 @@ function _chemGetDiversities(limit: number, molStringsColumn: DG.Column, fingerp
...
 
-  col.setTag(colNameTag, fingerprintColumnName);
-  if (createdMols)
-    col.setTag(FING_COL_TAGS.molsCreatedForVersion, String(col.version + 3));
-  col.setTag(colVerTag, String(col.version + 2));
-  col.setTag(FING_COL_TAGS.invalidatedForVersion, String(col.version + 1));
+  col.setTag(colNameTag, savedColumnName);
+}
+
+async function invalidateAndSaveColumns(molCol: DG.Column, fingerprintsType: Fingerprint,
```
skalkin, line 161: The "returnSmiles" looks like a code smell to me, can we make it simpler?

packages/Chem/src/rdkit-service/rdkit-service-worker-substructure.ts
```diff
@@ -29,95 +28,115 @@ export class RdKitServiceWorkerSubstructure extends RdKitServiceWorkerSimilarity
...
-          } else
-            mol.delete();
-        } // else, this looks to be a real SMARTS
-      } else { // failover to queryMolBlockFailover
-        // possibly get rid of fall-over in future
-        queryMol = getMolSafe(queryMolBlockFailover, {mergeQueryHs: true}, this._rdKitModule).mol;
+  searchWithSubstructLib(queryMol: RDMol): Uint32Array {
+    const matches = new BitArray(this._substructLibrary!.size() + this._malformedIdxs!.trueCount());
+    const str = this._substructLibrary!.get_matches(queryMol, undefined, undefined, this._substructLibrary!.size());
+    const matchesIdxs = JSON.parse(str);
```
skalkin, line 86: I will never believe that parsing JSON as an array is more efficient than simply passing Int32Array :) Show me the numbers! :)

packages/Chem/src/rdkit-service/rdkit-service-worker-substructure.ts
```diff
@@ -29,95 +28,115 @@ export class RdKitServiceWorkerSubstructure extends RdKitServiceWorkerSimilarity
...
+    // re-calculate matches idxs considering malformed data
+    let nonMalformedCounter = 0;
+    let matchesCounter = 0;
+    for (let i = !this._malformedIdxs!.getBit(0) ? 0 : this._malformedIdxs!.findNext(0, false); i !== -1; i = this._malformedIdxs!.findNext(i, false)) {
```
skalkin, line 90: Let's simplify the loop

packages/Chem/src/rdkit-service/rdkit-service-worker-substructure.ts
```diff
@@ -29,95 +28,115 @@ export class RdKitServiceWorkerSubstructure extends RdKitServiceWorkerSimilarity
...
+    return matches.buffer;
+  }
 
-    return '[' + matches.join(', ') + ']';
+  searchWithoutPatternFps(queryMol: RDMol): Uint32Array {
+    const matches = new BitArray(this._rdKitMols!.length);
+    for (let i = 0; i < this._rdKitMols!.length; ++i) {
+      try {
+        if (this._rdKitMols![i] && this._rdKitMols![i]!.get_substruct_match(queryMol) !== '{}')
+          matches.setBit(i, true);
```
skalkin, line 125: No need for the notification. And actually use the setFast method everywhere in these loops, it is faster and has less parameters

skalkin (approved):
LGTM, but please look at the comments and address some of them as well

packages/Chem/src/analysis/r-group-analysis.ts
```diff
@@ -87,12 +88,14 @@ export function rGroupAnalysis(col: DG.Column): void {
...
             for (let i = 0; i < resCol.length; i++) {
               const molStr = resCol.get(i);
+              let mol: RDMol | null = null;
               try {
-                const mol = module.get_mol(molStr);
+                mol = module.get_mol(molStr);
                 molsArray[i] = mol.get_molblock().replace('ISO', 'RGP');
-                mol.delete();
               } catch (e) {
                 console.warn(`RGroupAnalysisWarning: skipping invalid molecule '${molStr}' at index ${i}`);
```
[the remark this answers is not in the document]
MariaDolotova: Removed

packages/Chem/src/analysis/chem-similarity-viewer.ts
```diff
@@ -166,7 +166,9 @@ export class ChemSimilarityViewer extends ChemSearchBaseViewer {
   }
 
   isEmptyOrMalformedValue(): boolean {
```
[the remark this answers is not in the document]
MariaDolotova: Modified malformed handling. Get information about malformed when trying to create fp.

packages/Chem/src/chem-searches.ts
```diff
@@ -255,6 +296,27 @@ export async function chemSubstructureSearchLibrary(
   }
 }
 
+function getMoleculesFilteredByPatternFp(molStrings: (string | null)[], filteredMolsIdxs: BitArray): string[] {
+  const filteredMolecules = Array<string>(filteredMolsIdxs.trueCount());
+  let counter = 0;
+  for (let i = filteredMolsIdxs.getBit(0) ? 0 : filteredMolsIdxs.findNext(0); i !== -1; i = filteredMolsIdxs.findNext(i)) {
```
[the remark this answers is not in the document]
MariaDolotova: Done

packages/Chem/src/chem-searches.ts
MariaDolotova: Because data contains invisible columns which will be added to dataframe and for each added column we will add a tag containing column version. And the final column version depends on the number of added tags. 
For instance we are adding both pattern fp and canonical smiles. So we are adding 2 tags:
'.Pattern.version' and '.canonicalSmiles.Version'. Anyway, I'm now looking at all this logic with tags and I think this should be simplified. Will look into it once more

packages/Chem/src/chem-searches.ts
MariaDolotova: Not sure if we can get rid of this additional parameter. The situation is the following: when we calculate pattern fp we also might need to return canonical smiles (and I want to do this in the same function not to create mol objects repeatedly). But we also might not need to calculate canonical smiles at least in 2 cases:
- we are creating Morgan fp;
- we already have a column which contains canonical smiles (for instance when I extracted molecules from Chembl I aleady had a column with canonical smiles and did not want to spend time and memory to create additional column).

So I need this additional parameter to know if creating of canonical smiles is required. @skalkin , do you think there is a better way to do this?

packages/Chem/src/rdkit-service/rdkit-service-worker-substructure.ts
MariaDolotova: @skalkin , not sure I got your question correctly. get_matches() method of SubstrcuctLibrary returns matches numbers as an array of numbers in form of string (that is how it is implemented in rdkit). And the only way to get the numbers is to call JSON.parse.

packages/Chem/src/rdkit-service/rdkit-service-worker-substructure.ts
MariaDolotova: Done

packages/Chem/src/rdkit-service/rdkit-service-worker-substructure.ts
MariaDolotova: Changed to setFast where possible

PR #2146, opened 2023-07-11
GROK-135535: Chem | Descriptors for a cell widget

StLeonidas:
- calculation of descriptors for a single cell now uses grok compute
- descriptors app removed; reason - all the exposed features are now covered from the box for a column and for a cell


skalkin (commented):
It appears it's obsolete (and we don't use scripting for calculating descriptors). Am I right? If yes, please close the PR

PR #2266, opened 2023-08-06
GROK-13708: Bio | WebLogo: Optimize with splitterAsFastaSimple bio lib

tanas80:
SplitterFunc now returns ISeqSplitted allowing to return original string for single characters alphabet


tanas80:
Replaced with commit subject corrected

PR #2278, opened 2023-08-13
updated MinimalLib to 1.2.10 featuring log capture functionality

ptosco:
@MariaDolotova @StLeonidas This PR exposes log capture functionality. You may use it as follows:
```
let rdkitModule = chem.getRdKitModule();
let logHandle;
let mol;
try {
    logHandle = rdkitModule.set_log_capture("rdApp.*");
    mol = rdkitModule.get_mol("CN(C)(C)C");
    if (!mol) {
        const logBuffer = logHandle.get_buffer();
        console.log(logBuffer);
        logHandle.clear_buffer();
    }
} finally {
    if (logHandle) {
        logHandle.delete();
    }
    if (mol) {
        mol.delete();
    }
}
```
By replacing `set_log_capture` with `set_log_tee`, the message will also be logged to the browser developer console.
`rdApp.*` captures all RDKit logs; you may restrict capture to individual logs replacing `rdApp.*` with `rdApp.debug`, `rdApp.info`, `rdApp.warning`, `rdApp.error`.

PR #2288, opened 2023-08-17
Expose capturing log to MinimalLib

ptosco:
@MariaDolotova this is the same PR as #2278 but with an added workaround to avoid the infinite loop that you were observing.
This was _very_ painful to track down.
The problem is loosely connected to [#6527](https://github.com/rdkit/rdkit/pull/6527) (which is why it has showed up only recently), but the infinite loop actually happens elsewhere. I will track it down in the following days.
Meanwhile here's a stopgap solution that should work for the time being.
I will get back to you once a proper fix is available; this is now filed here: [#6640](https://github.com/rdkit/rdkit/issues/6640).


MariaDolotova:
@ptosco , thank you! Will merge and test the PR.

ptosco:
@MariaDolotova yesterday the issue was fixed in `master`. Later this evening I can issue a new PR if you prefer to wait and test a proper fix.

PR #2317, opened 2023-08-24
ComputeUtils: Tweak history input

andrey-sharapov-sc:


AufarZakiev (approved):
LGTM!
I think, I will further improve it using inheritance from `DG.InputBase`.
Practically, it should become `DG.InputBase<DG.FuncCall | null>`

PR #2340, opened 2023-09-07
GROK-13871: EDA: ANOVA

vmakarichev:
One-way analysis of variances is implemented.


packages/EDA/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # EDA changelog
 
+## 1.1.4 (WIP)
+
+This release concentrates on new statistical features. Analysis of variance (ANOVA) is provided.
```
skalkin, line 5: Make it as simple as possible. "Implemented ANOVA" is probably enough :)

packages/EDA/src/stat-tools.ts
```diff
@@ -0,0 +1,261 @@
...
+  sum: number,
+  sumOfSquares: number,
+  size: number,
+  data: number[] | undefined | null,
+};
+
+/** One-way ANOVA computation results. The classic notations are used (see [2], p. 290). */
+export type OneWayAnovaTable = {
+  /** sum of squares between groups */
+  SSbn: number,
```
skalkin, line 38: It violates field naming convention, I think `ssBn` would be better, unless there is a reason?

packages/EDA/src/stat-tools.ts
```diff
@@ -0,0 +1,261 @@
...
+  NON_EQUAL_FACTORS_VALUES_SIZE = 'non-equal sizes of factor and values arrays. INPUT ERROR.',
+  INCORRECT_SIGNIFICANCE_LEVEL = 'incorrect significance level. It must be from the interval (0, 1). INPUT ERROR.',
+  INCORRECT_SAMPLE_SIZE = 'incorrect size of sample. DATA FACTORIZAING ERROR.',
+};
+
+export type SampleData = {
+  sum: number,
+  sumOfSquares: number,
+  size: number,
+  data: number[] | undefined | null,
```
skalkin, line 32: Strongly-typed raw arrays are always preferred for number crunching. And why | undefined | null?

packages/EDA/src/stat-tools.ts
```diff
@@ -0,0 +1,261 @@
...
+  toCheckNormality: boolean = false, 
+  toCheckVariancesEqaulity: boolean = false): AnovaResults
+{
+  checkSignificanceLevel(alpha);
+
+  const values = ((valuesCol.type === DG.COLUMN_TYPE.FLOAT) || (valuesCol.type === DG.COLUMN_TYPE.INT)) 
+    ? valuesCol.getRawData() 
+    : valuesCol.toList();
+
+  const factors = factorsCol.toList();
```
skalkin, line 237: Why are we creating another copy of the list, can't we work with the column directly?

packages/EDA/src/stat-tools.ts
```diff
@@ -0,0 +1,261 @@
...
+
+  if (toCheckVariancesEqaulity)
+    if(!areVariancesOfFactorizedDataEqual(factorizedData))    
+      throw new Error('Variances are not equal!');
+  
+  const anovaTable = computeOneWayAnovaTable(factorizedData);
+
+  const Fcritical = jStat.centralF.inv(1 - alpha, anovaTable.DFbn, anovaTable.DFwn);
+
+  const conclusion = (anovaTable.Fstat > Fcritical)
```
skalkin, line 253: Let's simplify the wording.

skalkin (changes requested):

packages/EDA/src/stat-tools.ts
vmakarichev: Fixed!
The main reason is to follow notations given in the source [2, p. 290] :).

packages/EDA/src/stat-tools.ts
vmakarichev: For the case of **10M** rows, time of processing is:
- **312-497 ms** if we use a copy (const factors = factorsCol.toList() and then use factors[i]);
- **566-603 ms** if we work with a column directly.

What should be decreased in this case: time or spatial complexity?

packages/EDA/src/stat-tools.ts
vmakarichev: In ANOVA, the array `data ` is not used, just `sum`s are required. So, it can be undefined. Optionally, when splitting values with respect to factor levels (categories), data can be collected. For this, the function `getFactorizedData `should be used instead `getLevelsStat`.
Next, in this particular case, `data` is not specified as a strongly-typed raw array, since its size depends on the `factors` content.

skalkin: > Next, in this particular case, data is not specified as a strongly-typed raw array, since its size depends on the factors content.

Still it's inefficient. Perhaps we need `Column.splitByCategories(Column categorical): Map<String, Float32Array>` method?

packages/EDA/src/stat-tools.ts
skalkin: We should work with the column's getRawData, this is the best of both worlds, I think

packages/EDA/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+
+Implemented ANOVA.
+
+### Features
+
+* One-way ANOVA (see Top Menu > ML > Analysis of Variances (ANOVA)...)
+
 ## 1.1.3 (2023-08-24)
 
 This release is centered around enhancing user-friendliness and addressing concerns.
```
skalkin, line 13: Let's cut the BS next time :) If there is nothing specific to say but we need to put it (not a case here I think), it should be much simpler "Bug fixes and performance improvements".

skalkin (approved):

packages/EDA/src/stat-tools.ts
vmakarichev: I think that we do need this method!

PR #2345, opened 2023-09-12
ComputeUtils: RFV validation

andrey-sharapov-sc:
- Initial custom validation
- Better subscription management via rxjs in function views

### DIFF
--- a/libraries/compute-utils/function-views/src/rich-function-view.ts
+++ b/libraries/compute-utils/function-views/src/rich-function-view.ts
@@ -115,25 +115,21 @@ export class RichFunctionView extends FunctionView {
         return fieldValidations$.pipe(
           tap((payload) => this.setValidationPending(payload.field)),
           switchMap((payload) => {
-            return of(null).pipe(
-              // run revalidations immediately
-              payload.isRevalidation ? identity : delay(VALIDATION_DEBOUNCE_TIME),
-              mergeMap(() => {
-                const controller = new AbortController();
-                const signal = controller.signal;
-                let done = false;
-                return new Observable<Record<string, ValidationResult | undefined>>((observer) => {
-                  const sub = from(this.runValidation({...payload}, signal)).subscribe((val) => {
-                    done = true;
-                    observer.next(val);
-                  });
-                  return () => {
-                    if (!done)
-                      controller.abort();
-                    sub.unsubscribe();
-                  };
-                });
-              }),
+            const controller = new AbortController();
+            const signal = controller.signal;
+            let done = false;
+            const obs$ = new Observable<Record<string, ValidationResult | undefined>>((observer) => {
+              const sub = from(this.runValidation({...payload}, signal)).subscribe((val) => {
+                done = true;
+                observer.next(val);
+              });
+              return () => {
+                if (!done)
+                  controller.abort();
+                sub.unsubscribe();
+              };
+            });
+            return obs$.pipe(
               tap((results) => {
                 this.setValidationResults(results);
                 this.runRevalidations(payload, results);
@@ -149,7 +145,7 @@ export class RichFunctionView extends FunctionView {
     });
     this.subs.push(validationSub);
 
-    // waiting for validation after enter is pressed
+    // waiting for debounce and validation after enter is pressed
     const runSub = combineLatest([
       this.runRequests.pipe(
         switchMap(() => of(false).pipe(
@@ -158,11 +154,8 @@ export class RichFunctionView extends FunctionView {
         ))),
       this.validationUpdates.pipe(debounceTime(0)),
     ]).pipe(
-      filter(([needToRun]) => {
-        if (needToRun)
-          return this.isRunnable();
-        return false;
-      })).subscribe(() => this.doRun());
+      filter(([needToRun]) => needToRun && this.isRunnable())
+    ).subscribe(() => this.doRun());
     this.subs.push(runSub);
 
     // always run validations on start
@@ -253,7 +246,7 @@ export class RichFunctionView extends FunctionView {
     }));
   }
 
-  private keepInputs() {
+  private keepOutput() {
     return this.func?.options['keepOutput'] === 'true';
   }
 
@@ -930,7 +923,7 @@ export class RichFunctionView extends FunctionView {
   }
 
   private syncOnInput(t: InputVariants, val: DG.FuncCallParam, field: SyncFields) {
-    const sub = getObservable(t.onInput.bind(t)).subscribe(() => {
+    const sub = getObservable(t.onInput.bind(t)).pipe(debounceTime(VALIDATION_DEBOUNCE_TIME)).subscribe(() => {
       if (this.isHistorical.value)
         this.isHistorical.next(false);
 
@@ -968,7 +961,7 @@ export class RichFunctionView extends FunctionView {
   }
 
   private async runValidation(payload: ValidationRequestPayload, signal: AbortSignal) {
-    const inputName = payload ? payload.field : undefined;
+    const inputName = payload.field;
     const inputNames = this.getValidatedNames(inputName);
 
     const validationItems = await Promise.all(inputNames.map(async (name) => {
@@ -1055,7 +1048,7 @@ export class RichFunctionView extends FunctionView {
   }
 
   private hideOutdatedOutput() {
-    if (this.keepInputs())
+    if (this.keepOutput())
       return;
     this.outputsTabsElem.panes
       .filter((tab) => Object.keys(this.categoryToDfParamMap.outputs).includes(tab.name))
--- a/packages/LibTests/src/tests/richFunctionViewTests.ts
+++ b/packages/LibTests/src/tests/richFunctionViewTests.ts
@@ -83,7 +83,7 @@ category('RichFunctionView Inputs', async () => {
       const element = (input as any).input;
       element.dispatchEvent(new Event('input', {bubbles: true}));
     }
-    await delay(100);
+    await delay(300);
     await view.doRun();
     const expected = {
       'inputs': [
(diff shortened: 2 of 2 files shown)


packages/Compute/src/package.ts
```diff
@@ -227,3 +227,16 @@ export function modelCatalog() {
     }
   } else grok.shell.v = modelsView;
 }
+
+//name: RangeValidatorFactory
+//input: object params
+//output: object validator
+export function RangeValidatorFactory(params: any) {
```
AufarZakiev, line 234: Suggestion: 

1) Introduce implicit first argument of type `DG.FuncCall` for all validators. Smth like `RangeValidatorFactory(currentCall:DG.FuncCall, params: any)`
2) Pass current call of the function to each validator

This way, it will be possible to depend validation result on the dynamic context. E.g. code this:
```
 return (val: number) => {
    const { min } = params;
    const { max } = currentCall.inputs['myMaxValueComputedOnPreviousStep'];
    if (val < min || val > max) {
      // TODO: format results
      return { errors: [`Out of range [${min}, ${max}] value: ${val}`] };
    }
  }
```

libraries/compute-utils/function-views/src/rich-function-view.ts
```diff
@@ -174,7 +198,9 @@ export class RichFunctionView extends FunctionView {
   private getStandardButtons(): HTMLElement[] {
     const runButton = this.getRunButton();
     const runButtonWrapper = ui.div([runButton]);
-    ui.tooltip.bind(runButtonWrapper, () => runButton.disabled ? (this.isRunning.value ? 'Computations are in progress' : 'Some inputs are invalid') : '');
+    ui.tooltip.bind(
+      runButtonWrapper,
+      () => runButton.disabled ? (this.isRunning.value ? 'Computations are in progress' : this.getValidationMessage()) : '');
```
AufarZakiev, line 203: I am sure it is better to keep validation messages inside of tooltips of invalid inputs, too. They will be visible on any hover over input - this will increase user-friendliness

libraries/compute-utils/function-views/src/rich-function-view.ts
```diff
@@ -845,24 +879,69 @@ export class RichFunctionView extends FunctionView {
...
+    this.clearValidation();
+    for (const [k, v] of this.funcCall.inputs.entries()) {
+      const standardMsgs = nonNullValidator(v, this._funcCall!);
+      if (standardMsgs) {
+        this.syncValidationState(k, standardMsgs);
+        continue;
+      }
+      const customValidator = this.validators[k];
+      if (customValidator) {
+        const customMsgs = customValidator(v, this._funcCall!);
```
AufarZakiev, line 901: Could we introduce `await` here to support async validators?
Or am I missing smth important?

packages/Compute/src/package.ts
```diff
@@ -227,3 +227,16 @@ export function modelCatalog() {
     }
   } else grok.shell.v = modelsView;
 }
+
+//name: RangeValidatorFactory
+//input: object params
```
AufarZakiev, line 232: Such input types are not supported outside of JS functions - but I think, it is not a problem.
1. Python/Octave/etc, validators are generally bad idea - they could not be fast
2. If smth outside of JS functions will be required - we easily can call Python functions right from JS validation functions (ofc, if we support async validators)

Noting that to be sure that we are on the same page regarding this question

packages/Compute/src/package.ts
AufarZakiev: Ah, I see `context` in `Validator` itself. It is OK then

libraries/compute-utils/function-views/src/rich-function-view.ts
AufarZakiev: I think, we will get old waiting for this :) 
We could keep it as it is now and introduce tooltips later

packages/Compute/src/package.ts
AufarZakiev: Don't know...
This `object` type could be tricky to implement for all languages supported. 

Anyway, the main idea for now is "use JS validators everywhere you can"

AufarZakiev: >If smth outside of JS functions will be required - we easily can call Python/Octave/etc. functions right from JS validation functions (ofc, if we support async validators) 

Exactly!

libraries/compute-utils/function-views/src/rich-function-view.ts
```diff
@@ -879,6 +1055,8 @@ export class RichFunctionView extends FunctionView {
   }
 
   private hideOutdatedOutput() {
+    if (this.keepInputs())
```
AufarZakiev, line 1058: Hiding output tabs creates problems somewhere?
If you want to prevent UI "blinking", I suggest:

1. avoid adding one more tag
2. add ui.setUpdateIndicator for output tabs while output is recalculated

AufarZakiev: Don't get it.

If the output result is used to tweak inputs - why couldn't we wait for the function run to end? Then see outputs as usual, analyze them and tweak inputs

Btw, it should be called `keepOutputs`, I think

packages/Compute/package.json
```diff
@@ -13,7 +13,7 @@
     "directory": "packages/Compute"
   },
   "dependencies": {
-    "@datagrok-libraries/compute-utils": "^1.14.3",
+    "@datagrok-libraries/compute-utils": "../../libraries/compute-utils",
```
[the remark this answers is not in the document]
AufarZakiev: Still see no pros to use this workaround. 
It is better to keep config clean, than changing it before merging into master

libraries/compute-utils/function-views/src/rich-function-view.ts
```diff
@@ -780,89 +856,189 @@ export class RichFunctionView extends FunctionView {
...
   }
 
+  private injectInputBaseValidation(t: DG.InputBase) {
+    const validationIndicator = ui.div('', {style: {display: 'flex'}});
+    t.addOptions(validationIndicator);
+    function setValidation(messages: ValidationResultBase | undefined) {
+      while (validationIndicator.firstChild && validationIndicator.removeChild(validationIndicator.firstChild));
+      const icon = getValidationIcon(messages);
+      if (icon)
+        validationIndicator.appendChild(icon);
```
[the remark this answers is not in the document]
AufarZakiev: Don't see overcomplications with showing/hiding approach.
We can
1. create three icons in the very beginning (notif-ons, warnings, errors) 
2. show/hide them on each `setValidation` call (via CSS or directly)
3. replace the content of validation dialog right before the click

Custom inputs should properly override `addOptions` in this case - we will use DG.InputBase as an interface (which is not the best, but still).

libraries/compute-utils/function-views/src/rich-function-view.ts
AufarZakiev: Aaah, got it. 

Well, this case should be resolved in different way: we should mark the output as outdated (it is very important for consistency and not-confusing users), but should not hide it. Let's keep your solution for now - later I will introduce better approach and will remove the tag. 

Is it OK?

packages/Compute/package.json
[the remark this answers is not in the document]
AufarZakiev: OK, I see your point.
CI/CD will just skip npm publishing step, so it could be used as a protection against occasional publish to npm

libraries/compute-utils/shared-components/src/FuncCallInput.ts
```diff
@@ -15,12 +14,99 @@ export interface FuncCallInput<T = any> {
...
+export function makeRevalidation(revalidate: string[], context?: any, result?: ValidationResultBase): ValidationResult {
+  return {revalidate, context, ...result};
+}
+
+export interface ValidationInfo {
+  param: string,
+  funcCall: DG.FuncCall,
+  lastCall?:DG.FuncCall,
+  isRevalidation: boolean,
+  isNewOutput: boolean,
```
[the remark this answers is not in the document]
AufarZakiev: Oh, our functions get more and more complicated dataflows... :(

libraries/compute-utils/function-views/src/rich-function-view.ts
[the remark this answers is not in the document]
AufarZakiev: > Custom inputs should not override or implement addOptions in general 
Why? It is much more flexible and we should force input developers to do everything right. Where is no principal difference between implementing `addOptions` and `setValidation` methods. It is better to use InputBase as API, as we did it with its' `root`, `onInput` and other methods. It is how I see it strategically and it is important.

> It is more complicated, instead of having a fresh immutable tree, we need to mutate a part of it
If we keep the overall structure (input -> its' icon -> its' popover), mutations are fine. They do less job, they keep code simpler. This is not very important - I think, we can keep it and later I will show you how it could be done another way. In the end of the day, I may be wrong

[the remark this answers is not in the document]
AufarZakiev: I see your solution is more flexible - but I don't like this flexibility. It is much better to force input developers to provide proper `addOptions` method and be control visual inside of it by the code of RFV.

In other words, if you have several stringInputs in the column (as single custom input) - no problem, provide a proper place for indication icons. It could be vertical div, a square in the corner, anything. Technically speaking, `FuncCallInput` class, from my point of view, will be united with InputBase in the end. This would provide an unified interface for all inputs.

I feel adding such a flexibility is a way to get crazy because of too much customization possibilities. And again, the code becomes a mess. If u r 100% sure that it is right way - OK then, I had to reassure that you see these suspicious places.

AufarZakiev (approved):

Issue #2368, opened 2023-09-26
Sketcher | structure lookup service: fix the behavior


ahadzhanian:
- [ ] Provide an indicator for search
- [ ] We shouldn't transform search entry to smiles
- [ ] Fix behavior when, after the first search, the search field is empty, and after the second one, the search entry is transformed into smiles

ahadzhanian:
Chem 1.8.3 compatible with DG 1.17.0

PR #2376, opened 2023-10-02
#2368: Chem: added progress indicator, fix input behaviour

MariaDolotova:


js-api/src/chem.ts
```diff
@@ -175,6 +175,9 @@ export namespace chem {
       return this._sketcherTypeChanged;
     }
 
+    get calculating(): boolean {return this.loader.style.display == 'initial';}
```
skalkin, line 178: Are you setting it to "initial" somewhere? Because I think otherwise it's going to be null

js-api/src/chem.ts
```diff
@@ -261,12 +264,17 @@ export namespace chem {
...
-        extractor
-          .apply([new RegExp(extractor.options['inputRegexp']).exec(x)![1]])
-          .then((mol) => this.setMolecule(mol));
-      else
-        this.setMolecule(x);
+        if (extractor != null && !checkSmiles(x) && !isMolBlock(x)) {
+          this.calculating = true;
+          extractor
+            .apply([new RegExp(extractor.options['inputRegexp']).exec(x)![1]])
+            .then((mol) => {
```
skalkin, line 271: You forgot about handling the exception, which is quite likely in this case

skalkin (changes requested):

js-api/src/chem.ts
MariaDolotova: Hm... I used exactly the same approach as our substructure filter does. In filter we do not set initial anywhere and everything works as it should. So I guess initial is set somewhere in loader constructor.

js-api/src/chem.ts
MariaDolotova: I tried to wrap in try{}catch{}. But in case we are using extractors from some package (which is always the case) and this extractor throwed an exception, this exception has never been catched by this catch in chem.ts. Anyway, I will put try catch block back.

js-api/src/chem.ts
skalkin: Well, double-check it please. As you can see we already had an issue with the loader, this can be related.

js-api/src/chem.ts
skalkin: > But in case we are using extractors from some package (which is always the case) and this extractor throwed an exception, this exception has never been catched by this catch in chem.ts.

Not sure what you meant? We would not catch it here? Why?

js-api/src/chem.ts
```diff
@@ -264,17 +264,21 @@ export namespace chem {
...
 
-        if (extractor != null && !checkSmiles(x) && !isMolBlock(x)) {
-          this.calculating = true;
-          extractor
-            .apply([new RegExp(extractor.options['inputRegexp']).exec(x)![1]])
-            .then((mol) => {
-              mol ? this.setMolecule(mol) : this.setMolecule('');
-              this.calculating = false;
-            });
+        try{
```
skalkin, line 267: Please follow our coding conventions

js-api/src/chem.ts
```diff
@@ -264,17 +264,21 @@ export namespace chem {
...
+            extractor
+              .apply([new RegExp(extractor.options['inputRegexp']).exec(x)![1]])
+              .then((mol) => {
+                mol ? this.setMolecule(mol) : this.setMolecule('');
+                this.calculating = false;
+              });
+          }
+          else
+            this.setMolecule(x);
+        } catch {
```
skalkin, line 279: This makes very little sense. An exception would still leave this.calculating in the true state. You should do extractor.apply(...).then(...).finally(...) instead.

js-api/src/chem.ts
MariaDolotova: In loader constructor we set class 'grok-loader' on a div element. And initial state for display there is 'inline-block'.

skalkin: I still have a problem with it and still don't get how it works. The unassigned style is "", and not "initial". If anything, you should check for the opposite (custom style).

PR #2390, opened 2023-10-04
GROK-10599: Auto unskip test if done

ksumag:


dnillovna:
We need to implement it in Jenkins

PR #2404, opened 2023-10-13
GROK-14060: Utils: render lines on scatter plot

MariaDolotova:


libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -1,31 +1,27 @@
...
-    points: Uint32Array;
-    id: number;
-    color?: string;
-    directionCol?: string
-    direction?: LineDirection;
-    opacity?: number;
-    width?: number;
+import BitArray from './bit-array';
+
+export type ILines = {
```
skalkin, line 5: I suggest to rename it to ILineSeries, this is typically what such things are called

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -1,31 +1,27 @@
...
+    from: Uint32Array;
+    to: Uint32Array;
+    colors?: string[];
```
skalkin, line 8: Typically, you would have one color and one width per series, as well as other parameters (like drawing arrows, etc). So I suggest to do the following:

```
  color?: string;       // common color. Use [colors] if you need individual colors per line
  colors?: string[];   // line colors. If they are the same for the series, use [color] instead.
```

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -1,31 +1,27 @@
...
+    widths?: Uint32Array;
+    opacities?: Float32Array;
+    drawArrows?: BitArray;
 }
 
 export class ScatterPlotLinesRenderer {
     sp: DG.ScatterPlotViewer;
-    xAxis: string;
-    yAxis: string;
+    xAxisValues: any[];
```
skalkin, line 16: Why is it "any"? And why do we need it?

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -38,13 +34,15 @@ export class ScatterPlotLinesRenderer {
         this.sp.render(this.canvas.getContext('2d') as CanvasRenderingContext2D);
     }
 
-    constructor(sp: DG.ScatterPlotViewer, xAxis: string, yAxis: string, lines: ILineOpts[]) {
+    constructor(sp: DG.ScatterPlotViewer, xAxis: string, yAxis: string, lines: ILines) {
         this.sp = sp;
-        this.xAxis = xAxis;
-        this.yAxis = yAxis;
+        this.xAxisValues = this.sp.dataFrame!.columns.byName(xAxis).toList();
```
skalkin, line 39: Why do we need it?

skalkin: I got it. This is TERRIBLY inefficient. We don't need it. You just doubled the memory amount we use for storing x,y numbers.

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -1,31 +1,27 @@
...
-    direction?: LineDirection;
-    opacity?: number;
-    width?: number;
+import BitArray from './bit-array';
+
+export type ILines = {
+    from: Uint32Array;
+    to: Uint32Array;
+    colors?: string[];
+    widths?: Uint32Array;
```
skalkin, line 9: Width is float.

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -1,31 +1,27 @@
...
+    opacities?: Float32Array;
+    drawArrows?: BitArray;
```
skalkin, line 11: Same, a single bool is probably enough for now.

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -63,77 +61,69 @@ export class ScatterPlotLinesRenderer {
...
-            pair[0] = p1;
-            pair[1] = p2;
-            const pairStr = `${this.lines[i].points[0]}#${this.lines[i].points[1]}`;
-            if (!this.numberOfLinesPerPair[pairStr])
-                this.numberOfLinesPerPair[pairStr] = 1;
-            else
-                this.numberOfLinesPerPair[pairStr]++;
+    countLinesPerPair(): void {
+        for (let i = 0; i < this.lines.from.length; i++) {
+            const pairStr = this.getStringPair(this.lines.from[i], this.lines.to[i]);
```
skalkin, line 105: StringPair is inefficient, come up with a better solution. One possibility is to sort lines, this way we can count them very efficiently

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -63,77 +61,69 @@ export class ScatterPlotLinesRenderer {
...
 
     getStringPair(point1: number, point2: number): string {
-        return `${point1}#${point2}`;
+        let p1 = point1;
+        let p2 = point2;
+        if (point1 > point2) {
+            p1 = point2;
+            p2 = point1;
+        }
+        return `${p1}#${p2}`;
```
skalkin, line 117: Very inefficient. Strings are expensive.

skalkin (changes requested):

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Done

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Removed

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Removed

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Done

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Sorting is not that trivial in this case. At least because the arrow direction can be set through points indexes order. For example we have [1, 4] in 'from' array and [4, 1] in 'to' array. But those are for the same point pair. At the moment I start to render I need to know the following for each point pair:
1) are there multiple lines for this pair of point (in this case I start drawing the first line as a curve) to form smth like this
image
2) the index of the line for this particular pair (to calculate the height of the curve and the side where it will be drawn).

I have proposed a decision without string pairs - createMultiLinesIndices() method which is only called once in the constructor. And generates an array with required information (basically a Uint32Array with an index of the line in case there are multiple lines for the pair) @skalkin , please take a look.

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: please see the comment above

libraries/utils/src/render-lines-on-sp.ts
skalkin: Thanks. So we are creating at least an additional string for each line? I don't like it, there are more efficient ways to do that (for instance by using a bit array to indicate whether we already have lines connecting the points, and then keeping a list of such lines somewhere).

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -63,58 +70,66 @@ export class ScatterPlotLinesRenderer {
...
+            const pointTo = this.sp.worldToScreen(this.xAxisCol.get(pointToIdx), this.yAxisCol.get(pointToIdx));
+            const bX = pointTo?.x;
+            const bY = pointTo?.y;
+            if (aX && aY && bX && bY) {
+                const line = new Path2D();
+                this.paths[i] = line;
+                line.moveTo(aX!, aY!);
+                const color = this.lines.color ?? this.lines.colors?.[i] ? this.lines.colors?.[i] : '0,128,0';
+                const opacity = this.lines.opacity ?? this.lines.opacities?.[i] ? this.lines.opacities?.[i] : 1;
+                this.ctx.strokeStyle = `rgba(${color},${opacity})`;
```
skalkin, line 91: Why would we create this string in the inner-most loop, when most likely it's the same color and the same opacity?

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -63,58 +70,66 @@ export class ScatterPlotLinesRenderer {
...
+                if (this.multipleLinesIndices[i]) {
+                    midPoint = this.findPerpendicularPointOnCurve(this.multipleLinesIndices[i],
+                        pointFrom.x, pointFrom.y, midPoint.x, midPoint.y)
                 }
+                line.quadraticCurveTo(midPoint.x, midPoint.y, bX, bY);
+                if (this.lines.drawArrows ?? this.lines.drawArrowsArr?.getBit(i))
+                    this.canvasArrow(line, changedPoints ? aX : bX, changedPoints ? aY : bY, midPoint.x, midPoint.y);
+                this.ctx.beginPath();
+                this.ctx.stroke(line);
+                this.ctx.closePath();
```
skalkin, line 104: And here's the biggest performance improvement: you should batch line draws in one stroke() method. Go ahead and compare the performance (I'm also curious). And I think that to keep the code clean, this means we should probably abandon opacities, lines and widths on the series level. Alternatively, you can keep it, but it will make the code a bit more complex. 

https://web.dev/articles/canvas-performance

Or, actually you can have the best of both worlds if you do stroke() in the loop only if you have line-specific attributes defined.

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -63,58 +70,66 @@ export class ScatterPlotLinesRenderer {
...
-    }
-
-    getStringPair(point1: number, point2: number): string {
-        let p1 = point1;
-        let p2 = point2;
-        if (point1 > point2) {
-            p1 = point2;
-            p2 = point1;
+    createMultiLinesIndices(): void {
+        const arrayIdxs = new Array<number>(this.lines.from.length);
```
skalkin, line 120: Why are you using an Array and not a strongly-type, super-efficient Int32List?

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -34,15 +39,17 @@ export class ScatterPlotLinesRenderer {
...
-        this.xAxisValues = this.sp.dataFrame!.columns.byName(xAxis).toList();
-        this.yAxisValues = this.sp.dataFrame!.columns.byName(yAxis).toList();;
+        this.xAxisCol = this.sp.dataFrame!.columns.byName(xAxis);
+        this.yAxisCol = this.sp.dataFrame!.columns.byName(yAxis);
         this.lines = lines;
         this.canvas = this.sp.getInfo()['canvas'];
         this.ctx = this.canvas.getContext('2d') as CanvasRenderingContext2D;
         this.paths = new Array<Path2D>(this.lines.from.length);
-        this.countLinesPerPair();
+        this.multipleLinesIndices = new Uint32Array(this.lines.from.length);
```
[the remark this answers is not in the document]
skalkin: If they are counts (not indices), you should 
1) rename the field
2) use Uint8List, don't think we should support more than 256 lines per point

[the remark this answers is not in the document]
skalkin: "Indices" is grammatically correct, however in CS they are usually called "indexes". At least I switched some time ago :)

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -34,15 +39,17 @@ export class ScatterPlotLinesRenderer {
...
+    constructor(sp: DG.ScatterPlotViewer, xAxis: string, yAxis: string, lines: ILineSeries) {
         this.sp = sp;
-        this.xAxisValues = this.sp.dataFrame!.columns.byName(xAxis).toList();
-        this.yAxisValues = this.sp.dataFrame!.columns.byName(yAxis).toList();;
+        this.xAxisCol = this.sp.dataFrame!.columns.byName(xAxis);
+        this.yAxisCol = this.sp.dataFrame!.columns.byName(yAxis);
         this.lines = lines;
         this.canvas = this.sp.getInfo()['canvas'];
         this.ctx = this.canvas.getContext('2d') as CanvasRenderingContext2D;
         this.paths = new Array<Path2D>(this.lines.from.length);
```
skalkin, line 50: Oops, this is a huge memory waste, can't we simply use a function to determine whether a point belongs to a line? I'm sure we have it somewhere.

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -63,58 +70,66 @@ export class ScatterPlotLinesRenderer {
...
-        let p1 = point1;
-        let p2 = point2;
-        if (point1 > point2) {
-            p1 = point2;
-            p2 = point1;
+    createMultiLinesIndices(): void {
+        const arrayIdxs = new Array<number>(this.lines.from.length);
+        for (let i = 0; i < arrayIdxs.length; i++)
+            arrayIdxs[i] = i;
+        while (arrayIdxs.length) {
```
skalkin, line 123: You are dynamically resizing an array? Not sure what the idea is, but it's inefficient. Please come up with a better implementation.

skalkin (changes requested):

libraries/utils/src/render-lines-on-sp.ts
```diff
@@ -63,58 +70,66 @@ export class ScatterPlotLinesRenderer {
...
-                        this.canvasArrow(line, bX, bY, midPoint.x, midPoint.y);
-                    this.ctx.beginPath();
-                    this.ctx.stroke(line);
-                    this.ctx.closePath();
+                let midPoint: DG.Point = this.midPoint(pointFrom, pointTo);
+                if (this.multipleLinesIndices[i]) {
+                    midPoint = this.findPerpendicularPointOnCurve(this.multipleLinesIndices[i],
+                        pointFrom.x, pointFrom.y, midPoint.x, midPoint.y)
                 }
+                line.quadraticCurveTo(midPoint.x, midPoint.y, bX, bY);
```
skalkin, line 99: Are we always drawing a quadratic curve, even if it's a straight line? I'm suspecting this is a lot more computationally intensive than just drawing a straight line

libraries/utils/src/render-lines-on-sp.ts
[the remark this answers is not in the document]
MariaDolotova: Done

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Changed - creating strokeStyle inside loop only in case of individual line styles

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Changed - do stroke inside loop only in only in case of individual line styles.

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Removed Path2D objects at all. Manipulating only on canvas context.

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Changed - in case of straight lines now using ctx.lineTo()

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Removed dynamic resizing

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: Changed to BitArray

libraries/utils/src/render-lines-on-sp.ts
MariaDolotova: As it was agreed, doing stroke for each separate line because otherwise drawing becomes too slow.

PR #2423, opened 2023-10-18
GROK-13009: SimPKPD: Modeling via WebAutoSolver

vmakarichev:
PK-PD modeling is provided by the applying WebAutoSolver.


packages/SimPKPD/src/package.ts
```diff
@@ -2,80 +2,75 @@
...
-          ], 'ui-form'),
-        ], 'ui-form'),
-      ], 'ui-form'),
-    ])
-  ]);
-  v.box = true;
+//name: PK-PD
+//description: Two-compartment pharmacokinetic-pharmacodynamic (PK-PD) simulation.
+//tags: model
+//input: double dose = 10000.0 {units: um; caption: dose; category: Dosing} [Dosage.]
```
skalkin, line 20: There is no need for the period in the end in description, use it in full sentences only.

packages/SimPKPD/src/package.ts
```diff
@@ -2,80 +2,75 @@
...
+//input: int dosesCount = 10 {caption: count; category: Dosing; min: 1; max: 1000} [Number of doses.]
+//input: double doseInterval = 12 {units: h; caption: interval; category: Dosing; min: 0; max: 100} [Dosing interval.]
+//input: double _KAVal = 0.3 {units: ; caption: rate constant; category: PK parameters}
```
skalkin, line 23: Just wondering if the units were omitted on purpose?

packages/SimPKPD/src/package.ts
```diff
@@ -2,80 +2,75 @@
...
+//input: double _CLVal = 2.0 {units: ; caption: clearance; category: PK parameters}
+//input: double _V2Val = 4.0 {units: ; caption: central volume; category: PK parameters} [Central compartment volume.]
```
skalkin, line 25: would be great to add reasonable min/max to parameters, so that we can use clickers and sliders in inputs

packages/SimPKPD/src/pk-pd-tools.ts
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  return {
+    C2: DG.Column.fromFloat32Array(CENTR_CONC, C2),
+    C3: DG.Column.fromFloat32Array(PER_CONC, C3)
+  }
+}
+
+/** Get simulation function with respect to the model specified */
+function getSimFn(compartments: string): any {
```
skalkin, line 62: Return type should not be "any", TypeScript lets you precisely define function types.

packages/SimPKPD/src/pk-pd-tools.ts
```diff
@@ -0,0 +1,165 @@
...
+const EFF_T = 'eff(t)';
+export const EFFECT = 'Effect';
+const T_COL = 't';
+export const TIME = 'Time [h]';
+export const CENTR_CONC = 'Central concentration';
+const PER_CONC = 'Peripheral concentration';
+const POINTS_COUNTER = 100;
+
+/** Compute concentrations */
+function getConc(centrCol: DG.Column, periCol: DG.Column, V2: number, V3: number): {C2: DG.Column, C3: DG.Column} {
```
skalkin, line 36: I'm not sure why are we dong this - is this some sort of a demo dataset? Probably need some comment on top

packages/SimPKPD/src/pk-pd-tools.ts
```diff
@@ -0,0 +1,165 @@
...
+      await oneIntervalSimulation(compartments, 
+        initial, final, step,
+        _depotInitial, _centrInitial, _periInitial, _effInitial, 
+        _KAVal, _CLVal, _V2Val, _QVal, _V3Val, _KinVal, _KoutVal, _EC50Val),
+      true
+    );
+  }
+
+  // Rename columns
+  const timeCol = res.col(T_COL);
```
skalkin, line 149: I think it's would be much shorter and a lot more clear to just have 5 lines instead of 15 like that:

```
res.col(T_COL)!.name = TIME;
res.col(EFF_T)!.name = EFFECT;
```

skalkin: Or simply use the same constant; or, define mapping in the dictionary

packages/SimPKPD/wasm/help-panel.js
```diff
@@ -0,0 +1,27 @@
+// Help pannels
```
skalkin, line 1: Typo (panels)

skalkin (approved):
Looks good! Left a bunch of minor comments.

packages/SimPKPD/src/package.ts
vmakarichev: Fixed

packages/SimPKPD/src/package.ts
vmakarichev: By default, WebAutoSolver generates "units: ..." in all cases.
FIXED

packages/SimPKPD/src/package.ts
vmakarichev: Done

packages/SimPKPD/src/pk-pd-tools.ts
vmakarichev: DONE.
Output columns contain data that is evaluated outside wasm-computations, after the main results are obtained.

packages/SimPKPD/wasm/help-panel.js
vmakarichev: Fixed

packages/SimPKPD/src/pk-pd-tools.ts
vmakarichev: Done

packages/SimPKPD/src/pk-pd-tools.ts
vmakarichev: Fixed

PR #2517, opened 2023-11-13
GROK-14223: Help: Release notes1.17.0

ahadzhanian:


help/deploy/releases/release-history.md
```diff
@@ -37,374 +37,140 @@ See also:
 
 ## 2023-10-31 1.17.0
 
-### Addressed Issues
+Datagrok 1.17 release focuses on stability, performance, and usability improvements:
+
+* Initial Settings that enable you to configure all necessary options before using the platform.
```
lbankurova, line 42: Not clear. If "Initial Setting" is a UI element, it should be bold. If not, it shouldn't be capitalized. Also, what does it mean, "initial settings"? 
"settings that enable you" - too long
"all necessary options" can be replaced with one word -  settings
"before using the platform" - I thought it could be done at any time? 

Try this option: The ability to configure...using ...

PR #2518, opened 2023-11-13
GROK-12294: Added ability to send email using js-api

PavloPolovyi:


js-api/src/dapi.ts
```diff
@@ -381,6 +381,43 @@ export class AdminDataSource {
...
+    if (envelope.to.length === 0)
+      throw new Error('Recipients list shouldn\'t be empty');
+    return api.grok_Dapi_Admin_Send_Email(this.dart, envelope);
+  }
+}
+
+/**
+ * Represents message that can be sent over the network using the configured SMTP service
+ */
+export interface Envelope {
```
skalkin, line 400: Shouldn't we call it Email, or is Envelope a commonly used name in such cases?

PavloPolovyi: Yeah, I read a little about the semantics of these words, and an envelope represents a message that can be sent to only one recipient. In our case it is incorrect. I will change it to `Email`;

PR #2522, opened 2023-11-14
fix(DDIS-6462): Fonts are too large compared to the scaffold size

ptosco:
@MariaDolotova @StLeonidas This small PR is to normalize the scale of scaffolds retrieved from the sketcher, such that their bond length is the same that RDKit expects. This avoids downstream issues such as:
1. heteroatom fonts being too large compared to the size of the scaffold
2. possible distortions upon attempting to align molecules to such scaffolds

For instance, fonts that previously looked like this
image

are now properly scaled:
image


ptosco:
@MariaDolotova I have noticed this was not included in `Chem 1.8.6`: does this break any of your use cases?

MariaDolotova:
@ptosco , oh, I'm sorry, I somehow missed the PR. Will merge and test.

PR #2542, opened 2023-11-22
Disable molecule alignment on filtering, highlight only

ptosco:
@MariaDolotova @StLeonidas This PR disables molecule alignment on filtering as discussed during the DG Weekly meeting - we only turn on the highlighting.

- Disable molecule alignment on filtering, highlight only
- Fixed typo in variable name


MariaDolotova:
Hi @ptosco !
We are now working on the same issue but the thing is that we do not want to switch the alignment at all. We are going to allow user to choose whether he wants to switch alignment or highlight on/off when filtering. This is going to be implemented via additional checkboxes in sketcher itself. And the selected values are going to be user specific and stored in a user storage.

PR #2585, opened 2023-12-12
GROK-14019: Add js api for package settings

KiK0S:


js-api/src/entities.ts
```diff
@@ -1000,9 +1000,22 @@ export class Package extends Entity {
...
+    return this.getSettings();
+  }
+
+  /** Returns settings for a package. */
+  getSettings(): Promise<Map<string, any>> {
+    return api.grok_Package_Get_Settings(this.name);
+  }
+
+  /** Returns settings for a package. */
+  setSettings(props: Map<string, any>) {
```
alex-aprm, line 1017: Let's add group parameter

KiK0S: done

alex-aprm (approved):

PR #2592, opened 2023-12-18
GROK-14322: API samples: files in scripts

KiK0S:


packages/ApiSamples/scripts/scripting/files-for-scripting.js
```diff
@@ -0,0 +1,22 @@
...
+    let data = await fileInfo.readAsString();
+    grok.shell.info(data);
+}
+
+async function logFileContentAsBytes(fileInfo) {
+    let data = await fileInfo.readAsBytes();
+    grok.shell.info(data);
+}
+
+var line = 'test how script uses files'
```
KiK0S, line 15:
```suggestion
// these values are the same for utf-8
var line = 'test how script uses files'
```

PR #2648, opened 2024-01-24
GROK-14487: Diff Studio: Improvement

vmakarichev:
**Implemented features:**
* Tab control with `Model` and `Run` panes
* Interactive model exploration via the `Run` pane
* Routing of the example models
* Check correctness of model inputs annotation
* Comments (`//...`) in formulas
* Expressions usage in the `#output`-block
* Model *ivp*-files handling


packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,870 @@
...
+  EDITOR_STATE.ACID_PROD,
+  EDITOR_STATE.NIMOTUZUMAB,
+];
+
+/** Get problem with respect to IVP editor state */
 function getProblem(state: EDITOR_STATE): string {
   switch (state) {
-    case EDITOR_STATE.BASIC_TEMPLATE:
-      return TEMPLATES.BASIC;
+  case EDITOR_STATE.BASIC_TEMPLATE:
```
LaykDimon, line 51: maybe it's a minor detail. but switches almost always can be reduced
you can make an object like `{[key: string]: string}` (you can look at that detail in platform-viewers-demo.ts in the Tutorials package), and then use `arrayName[state]` to get the value
so you will have less code and less logic in this situation

LaykDimon: image

packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,870 @@
...
+    multiAxisLegendPosition: 'RightTop',
+    segmentColumnName: colNames.includes(STAGE_COL_NAME) ? STAGE_COL_NAME: null,
   };
 }
 
-/** Run solver application */
-export async function runSolverApp() {
+/**  String to value */
+const strToVal = (s: string) => {
+  const num = Number(s);
```
LaykDimon, line 155: to convert a number to a string you can simply use const num = +s;

packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,870 @@
...
-      await call.call();
+  if (s === 'true')
+    return true;
 
-      const sView = DG.ScriptView.create(script);
-      grok.shell.addView(sView);
+  if (s === 'false')
+    return false;
+
+  return s;
```
LaykDimon, line 166: I think you can make this strToVal function in one string - something like
`return !isNan(+s) ? num : s === 'true' ? true : s === 'false' ? false : s`

LaykDimon: and just make it 2 lines long for the clarity

packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,870 @@
...
+                const eqIdx = equality.indexOf(PATH.EQ);
+                this.startingInputs?.set(equality.slice(0, eqIdx).toLowerCase(), Number(equality.slice(eqIdx + 1)));
+              });
+            } catch (error) {
+              this.startingInputs = null;
+            }
+          }
+
+          await this.setState(model as EDITOR_STATE, false);
+        } else
```
LaykDimon, line 203: better start else on the next line

packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,870 @@
...
+  } // runSolverApp
+
+  /** Run Diff Studio demo application */
+  public async runSolverDemoApp(): Promise<void> {
+    this.createEditorView(DEMO_TEMPLATE, true);
+    this.solverView.setRibbonPanels([[this.openIcon, this.saveIcon, this.exportButton, this.helpIcon]]);
+    this.toChangePath = false;
+    const helpMD = ui.markdown(demoInfo);
+    const divHelp = ui.div([helpMD]);
+    divHelp.style.padding = '10px';
```
LaykDimon, line 217: can we move these styles to CSS?

packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,870 @@
...
 
-        if (toChangeSolutionViewerProps) {
-          solutionViewer.setOptions(lineChartOptions(solutionTable.columns.names()));
-          toChangeSolutionViewerProps = false;
+      saveBtn.hidden = false;
+    });
+
+    this.toRunWhenFormCreated = true;
+
+    setTimeout(() => {this.runSolving(true);}, TIMEOUT.PREVIEW_RUN_SOLVING);
```
LaykDimon, line 277: I think you can remove {} in here

packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,870 @@
...
-  let editorState: EDITOR_STATE = EDITOR_STATE.BASIC_TEMPLATE;
-  let toShowWarning = true;
-  let isChanged = false;
-
-  editorView.dom.addEventListener('keydown', async (e) => {
-    if (e.key !== HOT_KEY.RUN)
-      isChanged = true;    
-  });
+    this.editorView.dom.style.overflow = 'auto';
+    this.editorView.dom.style.height = '100%';
```
LaykDimon, line 418: same for these styles

packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,870 @@
...
-  /** Run model application */
-  const runModelApp = async () => {
+      this.isSolvingSuccess = true;
+    } catch (error) {
+      this.clearSolution();
+
+      if (error instanceof Error)
+        grok.shell.error(error.message);
+      else
+        grok.shell.error(ERROR_MSG.SCRIPTING_ISSUE);
```
LaykDimon, line 642: you can make it in a ternary way
`grok.shell.error(error instanceof Error ? error.message : ERROR_MSG.SCRIPTING_ISSUE);`

packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,870 @@
...
+        this.tabControl.currentPane = this.modelPane;
+    } catch (error) {
+      this.prevInputsNode = null;
+      this.tabControl.currentPane = this.modelPane;
+      this.clearSolution();
+
+      if (error instanceof Error)
+        grok.shell.error(error.message);
+      else
+        grok.shell.error(ERROR_MSG.UI_ISSUE);
```
LaykDimon, line 669: same here

packages/ODEs/src/scripting-tools.ts
```diff
@@ -594,26 +639,108 @@ function getMathArg(funcIdx: number): string {
...
+} // getCustomOutputLinesNoExpressions
+
+/** Return custom output lines: with expressions */
+function getCustomOutputLinesWithExpressions(ivp: IVP): string[] {
+  const res = [''];
+
+  res.push(SCRIPT.CUSTOM_OUTPUT_COM);
+
+  // 1. Solution raw data
+  res.push(`const ${ivp.arg.name}RawData = ${DF_NAME}.col('${ivp.arg.name}').getRawData();`);
```
LaykDimon, line 677: push is a very expensive operations, maybe I would try to make an assign by index, but it is up to you

packages/ODEs/src/app.ts
vmakarichev: Replace by the map object

packages/ODEs/src/app.ts
vmakarichev: Done

packages/ODEs/src/app.ts
vmakarichev: Done

packages/ODEs/src/app.ts
vmakarichev: Done

packages/ODEs/src/app.ts
vmakarichev: Done

packages/ODEs/src/app.ts
vmakarichev: Done

packages/ODEs/src/app.ts
vmakarichev: Done

packages/ODEs/src/scripting-tools.ts
vmakarichev: Sure, but it requires some time to figure out how to implement this

packages/ODEs/src/app.ts
skalkin: const num = +s;

Ouch. I would have never guessed what it does. I would suggest to probably avoid it for readability.

packages/ODEs/src/css/app-styles.css
```diff
@@ -0,0 +1,13 @@
+.demo-app-div-help {
```
skalkin, line 1: What is ".demo-app-div-help"? The package should not modify global styles. And if it's local, it should be prefixed with "eqs" or whatever the package prefix is

skalkin (approved):
Great work! Just a few minor comments from me.

packages/ODEs/src/scripting-tools.ts
vmakarichev: Currently, the largest is 193 (the FAE case)

packages/ODEs/src/css/app-styles.css
vmakarichev: Fixed

packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,846 @@
...
+      grok.shell.addTableView(this.solutionTable) :
+      DG.TableView.create(this.solutionTable, false);
+
+    this.solverView.helpUrl = LINK.DIF_STUDIO_REL;
+    this.solverMainPath = PATH.CUSTOM;
+    this.solutionViewer = null;
+    this.viewerDockNode = null;
+    this.toChangeSolutionViewerProps = false;
+    this.solverView.name = MISC.VIEW_DEFAULT_NAME;
+    this.toShowWarning = true;
```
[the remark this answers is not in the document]
vmakarichev: Done

packages/ODEs/src/app.ts
```diff
@@ -1,654 +1,846 @@
...
+  private modelPane: DG.TabPane;
+  private runPane: DG.TabPane;
+  private editorView: EditorView | undefined;
+  private openMenu: DG.Menu;
+  private openIcon: HTMLElement;
+  private saveIcon: HTMLElement;
+  private helpIcon: HTMLElement;
+  private exportButton: HTMLElement;
+
+  constructor(toAddTableView: boolean = true) {
```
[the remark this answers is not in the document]
vmakarichev: "toAddTableView" is also used twice: 
1) when creating view (first after `constructor(...)`); - I reduce this line
2) in the line 300, when deciding whether to make timeout.

PR #2650, opened 2024-01-29
GROK-14640: Chem: Fixed error on sketcher resize

MariaDolotova:


skalkin (approved):
Not sure why it's not a CSS rule. Otherwise looks fine to me (presumably works even better on high dpi monitors?). Feel free to merge

PR #2672, opened 2024-02-09
GROK-14701: Sens Analysis: Column names issue fix

vmakarichev:
Fixed column names issue (non-unique case)

### DIFF
--- a/libraries/compute-utils/function-views/src/sensitivity-analysis-view.ts
+++ b/libraries/compute-utils/function-views/src/sensitivity-analysis-view.ts
@@ -746,7 +746,7 @@ export class SensitivityAnalysisView {
 
     // add columns with fixed inputs & mark them as fixed
     for (const col of fixedInputs) {
-      col.name = `${col.name} (fixed)`;
+      col.name = funcEvalResults.columns.getUnusedName(`${col.name} (fixed)`);
       funcEvalResults.columns.add(col);
     }
 
@@ -929,7 +929,7 @@ export class SensitivityAnalysisView {
 
     // add columns with fixed inputs & mark them as fixed
     for (const col of fixedInputs) {
-      col.name = `${col.name} (fixed)`;
+      col.name = funcEvalResults.columns.getUnusedName(`${col.name} (fixed)`);
       funcEvalResults.columns.add(col);
     }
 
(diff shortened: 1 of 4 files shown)


libraries/compute-utils/function-views/src/sensitivity-analysis-view.ts
```diff
@@ -86,12 +86,12 @@ export class SensitivityAnalysisView {
     const analysisInputs = {
       analysisType: {
         input: ui.choiceInput(
-          'Method', ANALYSIS_TYPE.GRID_ANALYSIS, [ANALYSIS_TYPE.GRID_ANALYSIS, ANALYSIS_TYPE.RANDOM_ANALYSIS, ANALYSIS_TYPE.SOBOL_ANALYSIS],
+          'Method', ANALYSIS_TYPE.RANDOM_ANALYSIS, [ANALYSIS_TYPE.GRID_ANALYSIS, ANALYSIS_TYPE.RANDOM_ANALYSIS, ANALYSIS_TYPE.SOBOL_ANALYSIS],
```
AufarZakiev, line 89: It is better to remove "GRID_ANALYSIS" value at all from enum, since it is easy to forget to comment out it everywhere. E.g. it is still here in the choiceInput's list.

Commenting out here is too complex, it is much better to remove everything completely. If we will need to restore this type of analysis, we always can revert a commit.

libraries/compute-utils/function-views/src/variance-based-analysis/sa-outputs-routine.ts
```diff
@@ -136,3 +136,33 @@ export function getOutput(funcCalls: DG.FuncCall[], outputsSpecification: Output
...
+        inCol.name = `${inCol.name} (input)`;
+        outCol.name = `${outCol.name} (output)`;
+      }
+    })
+  });
+
+  return inputs.concat(outputs);
+};
+
+export function getDataFrameFromInputsOutputs(inputs: DG.Column[], outputs: DG.Column[]): DG.DataFrame {
```
AufarZakiev, line 153: Isn't it better to create input/output columns with proper names initially (before calling `getDataFrameFromInputsOutputs`) ? 
This method mutates input parameters, which is no good for understanding and debugging

AufarZakiev (changes requested):

libraries/compute-utils/function-views/src/sensitivity-analysis-view.ts
vmakarichev: GRID_ANALYSIS is not removed. Just input for type of the segment grid (since now normal, log, etc. are not implemented) - it's Andrew's suggestion. 
RANDOM_ANALYSIS is used by default - again, Andrew's suggestion.

AufarZakiev: Got it. My suggestion stays the same - it is better to remove code lines at all instead of commenting them out

vmakarichev: Ok

libraries/compute-utils/function-views/src/variance-based-analysis/sa-outputs-routine.ts
vmakarichev: Agree. But, actually, I don't know how to implement this in the most compact way. This routine is used in both Random & Sobol methods.

My proposition is: to modify function name and add comments

AufarZakiev: Proper solution requires efforts and time to be done.
There is no need in creating `DG.Column` objects till the resulting DF is created. They could be created in the very end of the `perform` method. Until then, you may use simple structures and constuct the DF and columns only once before actually returning the analysis result.

This all could be done in refactoring stage. If something urgent should be delivered, I am OK with this PR

vmakarichev: Thanks

PR #2690, opened 2024-02-16
basic monomer lib file mangaer UI + validation

ghost:


tanas80 (approved):
Too many singletons.
On the next iteration splitting the monomer lib files management functionality please keep the "monomer provider" subsystem in the Bio package, and separate only management functions, maintaining a single entry point to MonomerLib API for monomer consumers (allowing notification with events).

PR #2705, opened 2024-02-26
GROK-15061: Help : Diff Studio

vmakarichev:
Docs on Diff Studio are updated with respect to its new version release


help/compute/diff-studio.md
```diff
@@ -171,18 +168,32 @@ You can set new values for parameters and change values for functions.
 
 ![Multi-stage model - loop](pics/DiffStudio-loop.gif)
 
-Apply the `#update` feature to obtain acyclic simulation.
-Set the length of the new modeling stage in the mandatory `length` variable.
-Inside you can use any mathematical expression to modify functions and parameters,
-same as for the `#loop` section.
+### Multi-stage model
```
skalkin, line 171: Multistage is one word.

help/compute/diff-studio.md
```diff
@@ -171,18 +168,32 @@ You can set new values for parameters and change values for functions.
...
+
+The multi-stage model is a model where parameters and function values can vary for different
```
skalkin, line 173: Simplify this.

skalkin: Something like that:

In multistage model, parameters might change between different argument ranges (stages).

help/compute/diff-studio.md
```diff
@@ -171,18 +168,32 @@ You can set new values for parameters and change values for functions.
...
-  length = 6
-  y = 1
+#argument: t, 1-st stage
+  t0 = 0.01
+  t1 = 15
+  h = 0.01
 ```
 
-Use any valid expression, when specifying the updates of inputs.
+Add the `#update` block. Input name of the stage right after the colon. Set the duration of the new modeling stage in the next line using
```
skalkin, line 185: > Input name of the stage right after the colon.

Simplify this. And it's "Enter", not "input". And it should be even simpler anyway :)

skalkin (approved):
Looks good to me, but address my minor comments first. And also I would simplify the block about "update:". OK to merge after that.

help/compute/diff-studio.md
```diff
@@ -145,12 +145,9 @@ Set [tolerance](https://pythonnumericalmethods.berkeley.edu/notebooks/chapter19.
...
 ```
 
-### Multi-stage model
-
-The multi-stage model is a model where parameters and function values can vary for different
-ranges of argument values.
+### Cyclic process simulation
 
-Datagrok provides a special capabilities for multi-stage simulation.
+Datagrok provides a special capabilities for modeling cyclic processes and phenomena.
```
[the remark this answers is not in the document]
vmakarichev: Done

help/compute/diff-studio.md
vmakarichev: Done

help/compute/diff-studio.md
vmakarichev: I've reduced it

PR #2752, opened 2024-03-25
Update minimallib to version 1.2.20

ptosco:
This PR contains support for R group decomposition and multi-color highlights similar to this blog post:
https://greglandrum.github.io/rdkit-blog/posts/2021-08-07-rgd-and-highlighting.html

RDKIt MInimalLib was built out of this branch:
https://github.com/ptosco/rdkit/tree/rgd-highlights-as-in-blog-post

@MariaDolotova @StLeonidas To learn how to use it, please refer to this HTML page:
https://ptosco.github.io/rgd/rgd_demo.html

I also exposed the `assignChiralTypesFromMolParity` flag; use it as follows if you need to assign atom chiral tags from parity rather than from wedge bonds. Please note that this flag should **only be used if the molecule has no wedge bond information but it has atom parity information**. In fact, setting the flag will cause the atom parity flag information to override the bond wedge information. This means that if the molecule has bond wedge information but no parity information (as most molecules do) and this flag is set, all stereochemistry information will be lost.


MariaDolotova:
Hi @ptosco ! Do I understand correctly that molV2000 has stereo information only in bond block, so there is no way we can restore molV3000 with CFG flags in both atoms and bond blocks?

ptosco:
@MariaDolotova The molblock that you shared with me was V2000, and it had stereo information _only_ in the CTAB as parity flags. This is non-standard, since stereo information is supposed to be stored _in the bond block_ using wedged and hashed bonds (`1` up, `6` down in V2000, `CFG=1` up, `CFG=3` down in V3000).
The parity information in the V2000 block is ignored on read as per the BIOVIA specification:

image

and does not really need to be present.

What I did was exposing to JS is a function that allows to rescue stereo information from atom parity on read, which is usually discarded, and use it to set chiral tags, which are in turn used to determine a bond wedging scheme through which stereochemistry is visualized.

The RDKit SD writer does not include atom parity information on output neither in V2000 nor in V3000 format since, based on BIOVIA specifications, that information is supposed to be ignored on read anyway.

MariaDolotova:
@ptosco , I got it. Thank you for clarification!

PR #2795, opened 2024-04-23
GitHub actions: Speedup improvements

dnillovna:


dnillovna:
@VHlushchen take a look at the CFN template change, you missed the issues which were found by linter

PR #2843, opened 2024-05-06
Tools: Add package option to grok test

dnillovna:


tools/bin/commands/test.ts
```diff
@@ -50,15 +45,25 @@ export function test(args: TestArgs): boolean {
...
 
-  const packageData = JSON.parse(fs.readFileSync(path.join(curDir, 'package.json'), {encoding: 'utf-8'}));
-  if (packageData.name) {
-    process.env.TARGET_PACKAGE = utils.kebabToCamelCase(utils.removeScope(packageData.name));
-    console.log('Environment variable `TARGET_PACKAGE` is set to', process.env.TARGET_PACKAGE);
-  } else {
-    color.error('Invalid package name. Set the `name` field in `package.json`');
+  if (args.package) {
+    process.env.TARGET_PACKAGE = utils.kebabToCamelCase(args.package);
+  } else
```
LaykDimon, line 50: why is it on different lines?

LaykDimon: i mean, else and if, i guess it's the same construction

LaykDimon (approved):

tools/bin/commands/test.ts
dnillovna: Added the braces to if-else. Could you check, @LaykDimon ?

LaykDimon (approved):

PR #2857, opened 2024-05-14
GROK-15637: Help: Optimization

vmakarichev:
Added docs on the parameter optimization


skalkin (approved):
Looks good to me! I have some minor suggestions but I'll just edit it in-place after we merge it

PR #2858, opened 2024-05-14
GROK-15299: Chem: Save functions to projects with data sync

MariaDolotova:


packages/Chem/src/analysis/activity-cliffs.ts
```diff
@@ -6,12 +6,21 @@ import {ITooltipAndPanelParams} from '@datagrok-libraries/ml/src/viewers/activit
...
+import {BitArrayMetrics} from '@datagrok-libraries/ml/src/typed-metrics';
 
 const canvasWidth = 200;
 const canvasHeight = 100;
 
 const cashedData: DG.LruCache<String, any> = new DG.LruCache<string, string>();
 
+export type ActivityCliffsParams = {
+  molColName: string,
+  actColName: string,
```
skalkin, line 18: Don't think we need to shorten "activity" here

packages/Chem/src/tests/projects-tests.ts
```diff
@@ -0,0 +1,363 @@
...
+  await delay(10);
+  await dataFrameContainsColumns(colList);
+  if (viewerType)
+    await checkViewerAdded(viewerType);
+  if (additionalChecks)
+    await additionalChecks(tv);
+}
+
+async function runActivityCliffs(tv: DG.TableView): Promise<void> {
+  await DG.Func.find({ package: 'Chem', name: 'activityCliffs' })[0].prepare({
```
skalkin, line 169: Why are we using `.prepare.call` if we can do `apply` directly?

Also, what's the reason for calling this function dynamically vs simply calling a JS function?

packages/Chem/src/tests/projects-tests.ts
```diff
@@ -0,0 +1,363 @@
...
+}
+
+async function runToInchiKeys(tv: DG.TableView): Promise<void> {
+  await DG.Func.find({ package: 'Chem', name: 'addInchisKeysTopMenu' })[0].prepare({
+    table: tv.dataFrame,
+    molecules: tv.dataFrame.col('smiles'),
+  }).call(undefined, undefined, { processed: false });
+}
+
+async function runToInchi(tv: DG.TableView): Promise<void> {
```
skalkin, line 275: What is this? Why do we have all these wrappers that do nothing but call a function?

packages/Chem/src/tests/projects-tests.ts
```diff
@@ -0,0 +1,363 @@
...
+  await grok.dapi.tables.save(tableInfo);
+  await grok.dapi.views.save(layoutInfo);
+  await grok.dapi.projects.save(project);
+  const projId = project.id;
+  grok.shell.closeAll();
+  const p = await grok.dapi.projects.find(projId);
+  await p.open();
+}
+
+async function dataFrameContainsColumns(colArr: string[]): Promise<void> {
```
skalkin, line 331: > Promise< void >
...
> return false;

?

packages/Chem/src/tests/projects-tests.ts
```diff
@@ -0,0 +1,363 @@
...
+      if (!grok.shell.tv.dataFrame.col(colName)) {
+        col = colName;
+        return false;
+      }
+    }
+    return true;
+  }, getError(), 5000);
+}
+
+async function checkViewerAdded(viewerType: string): Promise<void> {
```
skalkin, line 347: Same

skalkin (changes requested):
See few comments below; my main concern is why did we write all these wrappers?

packages/Chem/src/tests/projects-tests.ts
MariaDolotova: awaitCheck condition returns false. The function itself returns void

packages/Chem/src/tests/projects-tests.ts
MariaDolotova: awaitCheck condition returns false. The function itself returns void

packages/Chem/src/tests/projects-tests.ts
MariaDolotova: If I remember correctly, I need the fucntion to be called with 'processed: false' parameter. But if I call function directly, processed parameter is set to true.

packages/Chem/src/analysis/activity-cliffs.ts
MariaDolotova: Done

PR #2882, opened 2024-05-29
Webcomponents: Initial 

andrey-sharapov-sc:
How it works:

1. webcomponents library contains implementation of components (a small set of them, most importantly wrappers for Viewer and Form)
2. package Webcomponents registers webcomponets during package autostart, should be loaded before any app using them.
3. webcomponents-vue libaray wraps webcomponents into vue3 components, it consumes only types of webcomponets, the implementation of webcomponents is resolved in runtime.
4. LibTest includes examples of both raw webcomponents usage and their vue wrappers, as well as a webpack config to deal with tsx (most typescript tools/editors will support tsx, unlike any kind of templates).


skalkin (approved):
Looks good to me!

AufarZakiev (approved):

PR #2897, opened 2024-06-10
GROK-15756: Diff Studio: Solver update

vmakarichev:
Implemented
1) the ROS3PRw and ROS34PRw methods for numerical solving ODEs
2) callback mechanism for solvers control
3) #meta.solver-block for solving method control:  method, max iterations count and max computation time
4) max computation time control feature


packages/ODEs/src/app.ts
```diff
@@ -576,6 +578,8 @@ export class DiffStudio {
 
   /** Solve IVP */
   private async solve(ivp: IVP, inputsPath: string): Promise<void> {
+    const noCustomSettings = (ivp.solverSettings === DEFAULT_SOLVER_SETTINGS);
```
skalkin, line 581: Better to avoid negation in naming, as this leads to confusion; I would have called it `customSettings`

packages/ODEs/src/app.ts
```diff
@@ -931,4 +946,69 @@ export class DiffStudio {
...
+      await call.call();
+    } catch (err) {
+      if (!(err instanceof CallbackAction))
+        throw new Error((err instanceof Error) ? err.message : ERROR_MSG.SCRIPTING_ISSUE);
+    }
+  }
+
+  /** Close previousely opened performance dialog */
+  private closePerformanceDlg(): void {
+    if (this.performanceDlg) {
```
skalkin, line 972: Probably simpler to just do that:

this.performanceDlg?.close();
this.performanceDlg = null;

packages/ODEs/src/app.ts
```diff
@@ -931,4 +946,69 @@ export class DiffStudio {
...
+      this.performanceDlg.close();
+      this.performanceDlg = null;
+    }
+  }
+
+  /** Show performance dialog */
+  private async showPerformanceDlg(ivp: IVP, inputsPath: string): Promise<void> {
```
skalkin, line 979: I don't see any async code inside, I think this method should be sync.
Also, it might probably be a good idea for it to return the dialog (to be able to control the result later)

packages/ODEs/src/scripting-tools.ts
```diff
@@ -1074,6 +1092,34 @@ function getSolutionDfColsNames(ivp: IVP): string[] {
...
+  const settings = new Map<string, string>();
+
+  const openBraceIdx = line.indexOf(BRACE_OPEN);
+  const closeBraceIdx = line.indexOf(BRACE_CLOSE);
+  let sepIdx: number;
+
+  if (openBraceIdx >= closeBraceIdx)
+    throw new Error(`${ERROR_MSG.BRACES}. Check the line '${line}'`);
+
+  line.slice(openBraceIdx + 1, closeBraceIdx).split(ANNOT_SEPAR).forEach((item: string) => {
```
skalkin, line 1106: I know it's minor, but still, let's prefer the canonical for loop instead of forEach, it's easier to read (also a little bit faster)

```
for (const item: string of  line.slice(openBraceIdx + 1, closeBraceIdx).split(ANNOT_SEPAR)) {
   ...
```

skalkin (approved):
Looks good to me! Left a couple of minor comments.

packages/ODEs/src/app.ts
vmakarichev: Fixed

packages/ODEs/src/app.ts
vmakarichev: Done

packages/ODEs/src/app.ts
vmakarichev: Corrected

packages/ODEs/src/scripting-tools.ts
vmakarichev: Done

PR #2923, opened 2024-07-01
GROK-16001: EDA: Softmax

vmakarichev:
Added softmax classifier


packages/EDA/src/package.ts
```diff
@@ -542,3 +546,213 @@ export function anova(table: DG.DataFrame, factor: DG.Column, feature: DG.Column
...
+//description: Linear regression demo
+//input: dataframe table
+//input: column_list features {type: numerical}
+//input: column target {type: numerical}
+//input: bool plot = true {caption: plot}
+export async function linearRegression(table: DG.DataFrame, features: DG.ColumnList, target: DG.Column, plot: boolean): Promise<void> {
+  const t1 = performance.now();
+  const params = await getLinearRegressionParams(features, target);
+  const t2 = performance.now();
+  console.log(`Fit: ${t2 - t1} ms.`);
```
skalkin, line 560: Let's not write to console in the production code

packages/EDA/src/package.ts
```diff
@@ -542,3 +546,213 @@ export function anova(table: DG.DataFrame, factor: DG.Column, feature: DG.Column
 export function kNNImputation() {
   runKNNImputer();
 }
+
+//name: Linear regression
+//description: Linear regression demo
```
skalkin, line 551: So is it a demo or not?

packages/EDA/src/package.ts
```diff
@@ -542,3 +546,213 @@ export function anova(table: DG.DataFrame, factor: DG.Column, feature: DG.Column
...
+//name: Generate linear regression dataset
+//description: Create demo dataset for linear regression
+//input: int rowCount = 10000 {min: 1000; max: 10000000; step: 10000}
+//input: int colCount = 10 {min: 1; max: 1000; step: 10}
+//input: double featuresScale = 10 {min: -1000; max: 1000; step: 10}
+//input: double featuresBias = 10 {min: -1000; max: 1000; step: 10}
+//input: double paramsScale = 10 {min: -1000; max: 1000; step: 10}
+//input: double paramsBias = 10 {min: -1000; max: 1000; step: 10}
+//output: dataframe table
+export function generateDatasetForLinearRegressionTest(rowCount: number, colCount: number,
```
skalkin, line 587: We have a bunch of demo datasets in `grok.data.demo.xxx`, do we need another one here? If we really do - perhaps we need to move it to grok.data.demo or to utils somewhere to be reused?

packages/EDA/src/package.ts
```diff
@@ -542,3 +546,213 @@ export function anova(table: DG.DataFrame, factor: DG.Column, feature: DG.Column
...
   runKNNImputer();
 }
+
+//name: Linear regression
+//description: Linear regression demo
+//input: dataframe table
+//input: column_list features {type: numerical}
+//input: column target {type: numerical}
+//input: bool plot = true {caption: plot}
+export async function linearRegression(table: DG.DataFrame, features: DG.ColumnList, target: DG.Column, plot: boolean): Promise<void> {
```
skalkin, line 556: Not sure what this function is doing - I mean, how is it making Datagrok better?
Does it modify the original table?

packages/EDA/src/package.ts
```diff
@@ -542,3 +546,213 @@ export function anova(table: DG.DataFrame, factor: DG.Column, feature: DG.Column
...
+//meta.mlname: Softmax
+//meta.mlrole: isApplicable
+//input: dataframe df
+//input: string predict_column
+//output: bool result
+export function isApplicableSoftmax(df: DG.DataFrame, predict_column: string): boolean {
+  for (const col of df.columns) {
+    if (col.name !== predict_column) {
+      if ((col.type !== DG.COLUMN_TYPE.INT) && (col.type !== DG.COLUMN_TYPE.FLOAT) && (col.type !== DG.COLUMN_TYPE.QNUM) && (col.type !== DG.COLUMN_TYPE.BIG_INT))
+        return false;
```
skalkin, line 697: But what's wrong if a dataframe contains a string column? We are not forced to use it as a predictor.
isApplicable is only supposed to check whether a model could be applicable to that dataset (essentially if a dataset contains the right number of feature/predictor columns). Right, @KiK0S ?

packages/EDA/src/package.ts
```diff
@@ -542,3 +546,213 @@ export function anova(table: DG.DataFrame, factor: DG.Column, feature: DG.Column
...
+  return (target.type === DG.COLUMN_TYPE.STRING);
+}
+
+//name: Softmax
+//desription: Softmax classifier demo
+//input: dataframe df
+//input: column_list features {type: numerical}
+//input: column target {type: string}
+//input: bool scatter = true
+export async function fitSoftmaxDemo(df: DG.DataFrame, features: DG.ColumnList, target: DG.Column, scatter: boolean) {
```
skalkin, line 718: I think we should come up with a better way for demo code, this just looks like a lot of code for such a simple purpose. Essentially all we need is a dataset, a model, and parameters, right? Perhaps there should be a helper method or something like this.

packages/EDA/src/softmax-classifier.ts
```diff
@@ -0,0 +1,390 @@
...
+        this.categories = new Array<string>(modelDf.rowCount);
+
+        for (let i = 0; i < c; ++i) {
+          const c = columns.byIndex(i);
+          this.categories[i] = c.name;
+
+          if (c.type !== DG.COLUMN_TYPE.FLOAT)
+            throw new Error('incorrect params columns type');
+
+          this.params[i] = c.getRawData() as Float32Array;
```
skalkin, line 97: Are we sure that the column is going to contains floats and not ints, for instance? Perhaps - I did not manage to trace it to the root

packages/EDA/src/softmax-classifier.ts
```diff
@@ -0,0 +1,390 @@
...
+        // extract params & categories
+        this.params = new Array<Float32Array>(c);
+        this.categories = new Array<string>(modelDf.rowCount);
+
+        for (let i = 0; i < c; ++i) {
+          const c = columns.byIndex(i);
+          this.categories[i] = c.name;
+
+          if (c.type !== DG.COLUMN_TYPE.FLOAT)
+            throw new Error('incorrect params columns type');
```
skalkin, line 95: Start the message with the capital letter. And it's "column type" not "columns type".
And I think more information is better. Like "Incorrect input column type. Expected: FLOAT, passed: STRING".

packages/EDA/src/svm.ts
```diff
@@ -494,23 +473,30 @@ function getUnpackedModel(packedModel: any): any {
...
-    res = res && (columns.byIndex(i).type == 'double' || columns.byIndex(i).type == 'int');
-  return res;
-}
\ No newline at end of file
+
+  if ((labels.type !== DG.COLUMN_TYPE.STRING) || (labels.categories.length > 2))
+    return false;
+
+  for (const col of columns) {
+    if ((col.type !== DG.COLUMN_TYPE.FLOAT) && (col.type !== DG.COLUMN_TYPE.INT))
```
skalkin, line 497: QNUM and BIGINT are not handled. I've seen it before somewhere, too. Let's use column.isNumerical?

packages/EDA/wasm/EDAAPI.js
```diff
@@ -129,3 +129,48 @@ export async function _trainAndAnalyzeLSSVMInWebWorker(gamma, kernel, kernelPara
...
+      resolve(getResult(EDA['fitLinearRegressionParamsWithDataNormalizing'], e.data));
+    }
+  });
+}
+
+export function _fitLinearRegressionParams(features, targets, paramsCount) {
+  return callWasm(EDA, 'fitLinearRegressionParams', [features, targets, paramsCount]);
+}
+
+export async function _fitLinearRegressionParamsInWebWorker(features, targets, paramsCount) {
```
skalkin, line 151: Oh nice, didn't even know we were doing it in the web worker! But linear regressions is very fast - it's up to you, but it's fine to do it in the main thread I think. I guess do it the way it makes sense to keep the code simple.

skalkin (approved):
Wow, that's a lot of code - impressive! Looks good to me, but please address my comments. Please also think if we can make it smaller (by reducing demo, or exported function)

packages/EDA/wasm/softmax.h
```diff
@@ -0,0 +1,154 @@
...
+					Z(i, j) = exp(Z(i, j)) * classWeights[i];
+			
+			sums = Z.colwise().sum();
+
+            // Compute predicted probabilities
+			for (int i = 0; i < m; ++i)
+				Z.col(i) = Z.col(i) / sums(i);			
+
+            // Copmute loss
+            loss = 0;
```
KiK0S, line 123: is everything okay with indentation here? was concerned about coomments as well but here is is a line of code that is out of its scope

KiK0S, line 123: is everything okay with indentation here? was concerned about comments as well but here is is a line of code that is out of its scope

packages/EDA/src/try.ts
KiK0S, line 1: i believe this should be excluded

packages/EDA/src/svm.ts
```diff
@@ -494,23 +473,30 @@ function getUnpackedModel(packedModel: any): any {
 export async function getPrediction(df: DG.DataFrame, packedModel: any): Promise<DG.DataFrame> {
   const model = getUnpackedModel(new Uint8Array(packedModel));
 
+  console.log(model);
```
KiK0S, line 476: do we need it?

packages/EDA/src/package.ts
vmakarichev: Removed this method

packages/EDA/src/package.ts
vmakarichev: It's a function for demo performance linear regression. REMOVED!

packages/EDA/src/package.ts
vmakarichev: Demo. Removed!

packages/EDA/src/package.ts
vmakarichev: Here, due to our predictive modeling features, df is a dataframe of column selected as features

packages/EDA/src/package.ts
vmakarichev: It's just a function for demo softmax performance. Removed!

packages/EDA/src/softmax-classifier.ts
vmakarichev: Modified

packages/EDA/src/svm.ts
vmakarichev: Fixed

packages/EDA/wasm/EDAAPI.js
vmakarichev: This file is generated automatically by our c++-to-wasm export script. For each exported c++-function, this script generates two versions: in web worker call and in the main thread. After this, we can choose the preferable approach.
Currently, in linear regression, we use the second one

packages/EDA/src/try.ts
vmakarichev: Removed!

packages/EDA/src/svm.ts
vmakarichev: Removed

packages/EDA/wasm/softmax.h
vmakarichev: can you clarify?

PR #2931, opened 2024-07-08
GROK-16231: EDA : PLS modeling

vmakarichev:
Added the PLS regression modeling feature


packages/EDA/src/package.ts
```diff
@@ -628,3 +556,58 @@ export function isApplicableLinearRegression(df: DG.DataFrame, predictColumn: DG
...
+  return DG.DataFrame.fromColumns([unpackedModel.predict(df.columns)]);
+}
+
+//name: isApplicablePLSRegression
+//meta.mlname: PLS Regression
+//meta.mlrole: isApplicable
+//input: dataframe df
+//input: column predictColumn
+//output: bool result
+export function isApplicablePLSRegression(df: DG.DataFrame, predictColumn: DG.Column): boolean {
```
KiK0S, line 596: Can you also add isInteractive? Format is the same and you can find examples in the master

packages/EDA/src/pls/pls-ml.ts
```diff
@@ -0,0 +1,378 @@
...
+  PARAMS = 'Regression Coefficients',
+  X_LOADING = 'x.loading.p',
+  LOADINGS = 'Loadings',
+  COMP = 'component',
+  COMPS = 'components',
+  EXPL_VAR = 'explained variance',
+  X_SCORE = 'x.score.t',
+  Y_SCORE = 'y.score.u',
+  SCORES = 'Scores',
+};*/
```
KiK0S, line 35: If we dont need this, i'd rather not push it

KiK0S (changes requested):
Looks good, left two minor comments and have a bit bigger question.

I see pls-tools that perform pls analysis, and it is good we reuse this code for the model. But for loadings and regression coefficients there is a special logic just for the model - can't we reuse the already existing one? Didn't quite catch where it is computed though - looks like in some webassembly.

packages/EDA/src/pls/pls-ml.ts
vmakarichev: Removed!

packages/EDA/src/package.ts
vmakarichev: Added

vmakarichev:
> Looks good, left two minor comments and have a bit bigger question.
> 
> I see pls-tools that perform pls analysis, and it is good we reuse this code for the model. But for loadings and regression coefficients there is a special logic just for the model - can't we reuse the already existing one? Didn't quite catch where it is computed though - looks like in some webassembly.

packages/EDA/src/package.ts
vmakarichev: Concerning the previous questions...
The main computations are performed at the wasm side (we also use them in pls-ml):
1) input: features (X) and target (Y)
2) output: regression coeffs, T,U-scores, P,Q-loadings (see https://en.wikipedia.org/wiki/Partial_least_squares_regression for more details), prediction
An important feature is that X, Y are normalized. So, regression coeffs DO NOT contain the bias term. In pls-ml, we need this term. For this reason, we perform another step - we compute it at the TS side.

Concerning loadings
For visualization purposes, we need loadings and explained variances. And they should be packed with regression coefs. For convenience of packing, we append each expl.variance to the corresponding loadings array.

So, in pls-ml, we use almost the same as in mulitvariate analysis in combination with some specific extras

KiK0S: ok, LGTM

KiK0S (approved):

PR #2935, opened 2024-07-16
Ssavenko/utils/benchmarkmode update

ssSavenko:


packages/ApiTests/src/functions/cache.ts
```diff
@@ -39,18 +39,18 @@ category('Benchmarks: Client-side cache', () => {
   test('Tiny scalar calls no cache', async () => {
     const iterations = DG.Test.isInBenchmark ? 100000 : 100;
     return DG.toDart(await runLoop(false, tiny, getTinyGenerator(true, iterations)));
-  }, {timeout: 400000});
+  }, {timeout: 400000, isEnabledBenchmarkMode: true});
```
alex-aprm, line 42: Please, rename to benchmark: true

PR #3003, opened 2024-08-22
grok tools: link: link update

ssSavenko:


tools/bin/commands/link.ts
```diff
@@ -1,175 +1,133 @@
...
-  });
+
+    if (anyChanges === false)
+      throw (new Error(`There is loop with next packages: ${JSON.stringify(Array.from(packagesToLink)).toString()}`));
+  }
+
+  let names = localPackageDependencies.map(x => x.name);
+  await runScript(`npm install`, curDir);
+  for (let name of names) {
+    await runScript(`npm link ${name}`, curDir);
```
drizhina, line 100: replace this with 
`await runScript('npm link ${names.join(' ')}', curDir);`

what you have can cause weird behavior by linking only last package, sometimes npm does weird things

drizhina (changes requested):

tools/bin/commands/link.ts
```diff
@@ -1,175 +1,133 @@
...
-  for (const dep in json.dependencies) {
-    if (dep === apiPackageName || dep.startsWith(`${libScope}/`))
-      libs[dep] = json.dependencies[dep];
+function collectPackagesData(packagePath: string = curDir): string[] {
+  const packageJsonPath = path.join(packagePath, 'package.json');
+  const json = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'));
+  let result: string[] = [];
+  let dependencies = json.dependencies;
+  if (devMode)
+    dependencies = json.devDependencies;
```
drizhina, line 46: I think it would be much better to link both in this case. I don't see a use case for linking only dev dependencies

drizhina (changes requested):

PR #3054, opened 2024-09-13
API: Utils: Added executeTests method

ssSavenko:


js-api/src/entities.ts
```diff
@@ -1125,6 +1127,137 @@ export class Package extends Entity {
       this._files = new FileSource(`System:AppData/${this.name}`);
     return this._files;
   }
+
+  private testsWasRegistered: boolean = false;
+  private packageTestFileLoaded: boolean = false;
+
+  private addNamespace(s: string, f: Func): string {
```
drizhina, line 1134: this function was self descriptive in test.ts file, but when its in js-api, it can be very easily confused with smth else. change the name to something that it actually does.

drizhina (approved):

PR #3064, opened 2024-09-17
GROK-16667: Help: Diff Studio

vmakarichev:
Added help on the value lookup feature


help/compute/diff-studio.md
```diff
@@ -424,6 +424,25 @@ By default, Diff Studio uses ROS34PRw and alerts you if computations take too lo
 
 Set the maximum number of iterations to debug formulas in complex models.
 
+## Lookup tables
+
+Lookup tables let you initialize model inputs with a set of pre-defined values. These dataframes specify names of inputs in the first column and their values in other numerical columns:
```
lbankurova, line 429: Can you simplify this? It describes a simple concept in a very complex way. How about this?

Lookup tables are pre-defined sets of model input values. They're organized as follows:

[Table]
To use a lookup table:
* Create a CSV file with your table and add it to  your project
* Add this [line/parameter] to your model:

```python code
```

Use the interface to select inputs and compare model runs:

[GIF]

lbankurova (changes requested):

help/compute/diff-studio.md
vmakarichev: Thanks! It's much more clear!

PR #3108, opened 2024-10-16
opalamarchuk/GROK-16787-Implement-correct-marker-sizes-for-ScatterPlot-WebGPU-rendering

Polaringu:
Added checks and fallbacks for WebGPU methods. Now when we had an error, we'll automatically switch to CPU processing.

In scatterplot WebGPU rendering/hit test, added multimple marker sizes support, using the texture atlas with pixel perfect textures for unblurred rendering. Added error throwing and also WGSL error reading for correct fallback mechanism.


packages/PowerGrid/src/package.ts
```diff
@@ -253,19 +260,50 @@ export function demoCellTypes() {
...
+//output: bool result
+export function isWebGPURenderValid(sc: DG.ScatterPlotViewer) {
+  if (scOriginalMarkerType == '')
+    scOriginalMarkerType = sc.props.markerType;
+  if (scOriginalMarkersColumnName == '')
+    scOriginalMarkersColumnName = sc.props.markersColumnName;
+  return sc.props.zoomAndFilter != 'pack and zoom by filter' 
+    && sc.props.markerType == scOriginalMarkerType
+    && sc.props.markersColumnName == scOriginalMarkersColumnName
+    && !sc.dataFrame.selection.anyTrue;
```
drizhina, line 304: I don't like this, I know I told you this idea, but now that I think of it, if scatterplot is preconfigured with marker col, and then the flag is turned on, this will not work. for now, as we only work with circles. the name of shape is circle, so just harcode the value of sc.props.markerType to be equal to DG.MARKER_TYPE.CIRCLE.
same with markerColumnName. make sure it is empty or empty string. `!sc.props.markersColumnName` will deal with both cases. and lose the global variables scOriginalMarkersColumnName and scOriginalMarkerType

drizhina (changes requested):
LGTM! one small comment to fix

drizhina (approved):

PR #3115, opened 2024-10-23
GROK-16923: implement web gpu colors support

Polaringu:
Added coloring for textures and dot types for scatterplot.
Implemented color caching and cache invalidation checks.
Introduced benchmarks for gpuRendering and gpuHitTest (see screenshot results on my PC with integrated AMD Radeon GPU on AMD Ryzen 7 7730U CPU)

image


packages/PowerGrid/src/tests/pinned-column-tests.ts
```diff
@@ -4,7 +4,10 @@ import * as DG from 'datagrok-api/dg';
 
 import $ from 'cash-dom';
 
-import {awaitCheck, category, expect, expectArray, test} from '@datagrok-libraries/utils/src/test';
+import {awaitCheck, category, expect, expectArray, test, testEvent} from '@datagrok-libraries/utils/src/test';
+import { _scWebGPURender } from '../package';
+
+export const _package = new DG.Package();
```
drizhina, line 10: nitpicking but please remove these :D

packages/PowerGrid/src/tests/webgpu-tests.ts
```diff
@@ -0,0 +1,53 @@
+import * as grok from 'datagrok-api/grok';
+import * as ui from 'datagrok-api/ui';
+import * as DG from 'datagrok-api/dg';
+
+import $ from 'cash-dom';
+
+import {awaitCheck, category, expect, expectArray, test, testEvent} from '@datagrok-libraries/utils/src/test';
+import { _scWebGPURender, _scWebGPUPointHitTest } from '../package';
```
drizhina, line 8: generally, this is a recipe for cyclic dependency errors :D 

instead of these, use the native grok func calls and remove imports from package:

const renderFunc = DG.Func.Find({tags: ['webGPURenderTag']})[0]
const hitTestFunc = DG.Func.Find({tags: ['webGPUHitTestTag']})[0]

and then use the arguments in structured way (yes the typing is lost, but build will be smaller and you will avoid any cyclic dependencies)
await renderFunc.apply({sc: sc, show: true})

packages/PowerGrid/src/tests/webgpu-tests.ts
```diff
@@ -0,0 +1,53 @@
...
+        const sp: DG.ScatterPlotViewer = view.scatterPlot();
+        sp.props.markerType = types[type];
+  
+        for (let i = 1; i <= 3; ++i) {
+          await DG.timeAsync(`${i} GPU Render ${size / 1000000.0}m ${types[type]}s`, async() => {
+            await _scWebGPURender(sp, true);
+          });
+        }
+      }
+    }
```
drizhina, line 30: you also need to specify that this test is benchmark, otherwise it will always run on tests and might fail a lot in CI/CD. the third argument in test function is an object where you can specify that:
image

packages/PowerGrid/src/webgpu/scatterplot.ts
```diff
@@ -22,6 +22,18 @@ class WebGPUCache {
   markerSizesLength = -1;
   markerDefaultSize = -1;
   sizeColumnName = '';
+  colorBuffer: GPUBuffer | null = null;
+  colorColumnName = '';
+  colorLength = -1;
+  colorAxisType: keyof typeof DG.AxisType = DG.AxisType.logarithmic;
```
drizhina, line 28: do I get this right that you initialize it to log because you expect a missmatch when linear comes by default?

drizhina (changes requested):

drizhina (approved):

drizhina (approved):

PR #3155, opened 2024-11-13
Help: Update help for predictive modeling

KiK0S:
* Interactive modeling
* Actualize model engines
* Add modeling visualizations


skalkin (approved):
Looks good to me, and great job! I will make some changes afterwards (streamlining and clarity), but the main idea is expressed the way we need it.

PR #3235, opened 2025-01-30
GROK-17352: Diff Studio: Facet grid plot

vmakarichev:
Added facet grid plot visualization


packages/DiffStudio/src/utils.ts
```diff
@@ -297,3 +297,50 @@ export function getCategoryWidget(category: string, inputs: DG.InputBase[]) {
...
+  case 18:
+    return 6;
+
+  default:
+    return 4;
+  }
+}
+
+/** RGB color to number */
+export function rgbToNum(rgbString: string) {
```
skalkin, line 330: Return type is missing. Also, we have a bunch of functions like that in the Color class, check it out (or add there if something is missing)

packages/DiffStudio/src/app.ts
```diff
@@ -2029,4 +2067,80 @@ export class DiffStudio {
...
+  /** Remove FacetGrid visualization */
+  private removeFacetGrid() {
+    if (this.facetGridNode && this.facetGridDiv) {
+      grok.shell.dockManager.close(this.facetGridNode);
+      this.facetGridDiv = null;
+    }
+  }
+
+  /** Return div with facet grid plot */
+  private getFacetPlot() {
```
skalkin, line 2080: Return type is missing

packages/DiffStudio/scripts/facet-layout.js
```diff
@@ -0,0 +1,77 @@
+//name: FacetPlot
```
skalkin, line 1: I wonder if we really need this script? Or is this for debugging purposes?

skalkin (approved):

packages/DiffStudio/scripts/facet-layout.js
vmakarichev: It is for debugging purposes.
Deleted

packages/DiffStudio/src/app.ts
vmakarichev: Fixed

packages/DiffStudio/src/utils.ts
vmakarichev: Removed this function. 
Coloring uses  `DG.Color.categoricalPalette` now

PR #3336, opened 2025-04-11
BatchCellRenderer: cell renderer update

ssSavenko:


libraries/utils/src/batch-cell-renderer.ts
```diff
@@ -0,0 +1,63 @@
...
+        this.cache.set(key, value);
+      }
+
+      this.cellsToLoad.forEach((x) => {
+        if (!this.loadedValues.has(x.cell.valueString))
+          this.cache.set(x.cell.valueString, null);
+        x.grid.invalidate();
+      })
+
+      this.loadedValues.clear();
```
drizhina, line 54: why do you need an instance variable for loadedValues? its used only in this closure

drizhina (changes requested):

libraries/utils/src/batch-cell-renderer.ts
ssSavenko: I need it for collecting loaded keys(not each would be loaded). I agree that it should not be instance variable moved it usual let

PR #3367, opened 2025-05-08
Ssavenko/decorators/curves update

ssSavenko:


packages/Curves/package.json
```diff
@@ -9,17 +9,17 @@
   },
   "dependencies": {
     "@datagrok-libraries/statistics": "^1.7.0",
-    "@datagrok-libraries/utils": "^4.5.7",
+    "@datagrok-libraries/utils": "../../libraries/utils",
     "@types/wu": "^2.1.42",
     "cash-dom": "^8.0.0",
-    "datagrok-api": "^1.25.0",
+    "datagrok-api": "C:\\Users\\sssav\\source\\repos\\datagrok\\reddata\\public\\js-api",
```
LaykDimon: please remove this from the path, it should be `../../js-api`

packages/Curves/src/package-functions.ts
```diff
@@ -0,0 +1,257 @@
+import * as DG from 'datagrok-api/dg';
+import * as grok from 'datagrok-api/grok';
+import * as ui from 'datagrok-api/ui';
+
+import { FitGridCellHandler, calculateSeriesStats, getChartDataAggrStats } from './fit/fit-grid-cell-handler';
```
LaykDimon, line 5: there shouldn't be any spaces after and before the parentheses

packages/Curves/src/package-functions.ts
```diff
@@ -0,0 +1,257 @@
...
+      });
+    df.columns.insert(column, chartColumn.idx);
+  }
+
+  //name: platesFolderPreview
+  //tags: folderViewer
+  //input: file folder
+  //input: list<file> files
+  //output: widget res
+  @grok.decorators.folderViewer({})
```
LaykDimon, line 109: I guess there is no need for annotation here as it was just to compare them before?

packages/Curves/src/package-functions.ts
```diff
@@ -0,0 +1,257 @@
...
+  //meta.browsePath: Plates
+  //output: view result
+  @grok.decorators.app({ name: 'Browse', browsePath: 'Plates'})
+  static platesApp() : any {
+    return platesAppView();
+  }
+
+  //input: dynamic treeNode
+  //input: view browseView
+  @grok.decorators.func({ })
```
LaykDimon, line 195: we need a decorator for this as well

packages/Curves/package.json
ssSavenko: It already fixed

PR #3378, opened 2025-05-22
MSA Header with interactive weblogo and "conservation tracks

rtviii:
Ok, first PR, be gentle.

Additions are basically thus (most you have seen, others mostly cosmetic):
- conservation track 
- weblogo track 
- tooltips 
- onHover events for weblogos
- row selection from a monomer's weblogo

Slightly refactored the whole thing with the constants at the top: lmk if it seems distasteful, i'm not hellbent on it, but frankly get lost doing arithmetic already with all the little pixel offsets for drawing & such, so trying to at least keep them within the same 50 lines seems sane. Performance seems good.

### DIFF
--- a/packages/Bio/src/widgets/sequence-scrolling-widget.ts
+++ b/packages/Bio/src/widgets/sequence-scrolling-widget.ts
@@ -232,11 +232,10 @@ export function handleSequenceHeaderRendering() {
               console.log('WebLogo track: no monomerLib available, using fallback colors');
 
 
-            webLogoTrack.setupDefaultTooltip(); // Enable tooltips
+            webLogoTrack.setupDefaultTooltip();
             tracks.push({id: 'weblogo', track: webLogoTrack, priority: 2}); // Higher visibility priority (2)
           }
 
-          // Create MSAScrollingHeader with initial height
           const scroller = new MSAScrollingHeader({
             canvas: grid.overlay,
             headerHeight: initialHeaderHeight,
--- a/packages/Bio/src/widgets/sequence-scrolling-widget.ts
+++ b/packages/Bio/src/widgets/sequence-scrolling-widget.ts
@@ -4,11 +4,11 @@ import * as grok from 'datagrok-api/grok';
 import * as DG from 'datagrok-api/dg';
 import * as ui from 'datagrok-api/ui';
 
-import { ConservationTrack, MSAHeaderTrack, MSAScrollingHeader, WebLogoTrack } from '@datagrok-libraries/bio/src/utils/sequence-position-scroller';
-import { MonomerPlacer } from '@datagrok-libraries/bio/src/utils/cell-renderer-monomer-placer';
-import { ALPHABET, TAGS as bioTAGS, SplitterFunc } from '@datagrok-libraries/bio/src/utils/macromolecule';
-import { _package } from '../package';
-import { ISeqSplitted } from '@datagrok-libraries/bio/src/utils/macromolecule/types';
+import {ConservationTrack, MSAHeaderTrack, MSAScrollingHeader, WebLogoTrack} from '@datagrok-libraries/bio/src/utils/sequence-position-scroller';
+import {MonomerPlacer} from '@datagrok-libraries/bio/src/utils/cell-renderer-monomer-placer';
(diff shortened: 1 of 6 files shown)


packages/Bio/src/widgets/sequence-scrolling-widget.ts
```diff
@@ -51,93 +167,170 @@ export function handleSequenceHeaderRendering() {
         const split = sh.splitter(seq);
         const maxSeqLen = split ? split.length : 30;
 
-        // makes no sense to have scroller if we have shorter than 50 positions
+        // Skip if sequence is too short
         if (maxSeqLen < 50)
           continue;
```
drizhina, line 172: I removed this in one of my commits, lets not skip short sequences, but rather not make the header expand if they are short, while leaving the ability to user to manually expand it and see the goodness

drizhina (changes requested):
Overall, very nice! but some major changes needed. most important is the lazy calculation of weblogo and conservation

drizhina (changes requested):

packages/Bio/src/widgets/sequence-scrolling-widget.ts
```diff
@@ -1,147 +1,740 @@
...
-          const cellBounds = e.bounds;
-          if (!cellBounds || cellBounds.height <= 50) {
-            scroller.headerHeight = 0;
-            return;
+
+        // Store the subscription for cleanup if needed
+        if (!grid.temp) grid.temp = {};
+        if (!grid.temp.filterSubs) grid.temp.filterSubs = [];
+        grid.temp.filterSubs.push(filterChangeSub);
+
```
drizhina, line 575: then you will not need to handle unsubscription or stuff like that, its automatically handled by the grid

packages/Bio/src/widgets/sequence-scrolling-widget.ts
```diff
@@ -1,147 +1,740 @@
...
-      handleGrid(grid);
+    if (grid) handleGrid(grid);
+  }
+}
+
+// Helper function to get visible sequence count
+function getVisibleSequenceCount(seqCol: DG.Column): number {
+  const dataFrame = seqCol.dataFrame;
+  if (dataFrame.filter.trueCount === dataFrame.rowCount)
+    return seqCol.categories.filter(Boolean).length;
```
drizhina, line 722: same here

packages/Bio/src/widgets/sequence-scrolling-widget.ts
```diff
@@ -1,147 +1,740 @@
...
+    if (dataFrame.filter.get(i)) {
+      const seq = seqCol.get(i);
+      if (seq) visibleSequences.add(seq);
+    }
+  }
+  return visibleSequences.size;
+}
+
+// Optional: Add cleanup method for when grids are destroyed
+export function cleanupMSAHeaderSubscriptions(grid: DG.Grid) {
```
drizhina, line 735: this will not be needed at all

packages/Bio/src/widgets/sequence-scrolling-widget.ts
```diff
@@ -1,147 +1,740 @@
...
+  }
+}
+
+// Helper function to get visible sequence count
+function getVisibleSequenceCount(seqCol: DG.Column): number {
+  const dataFrame = seqCol.dataFrame;
+  if (dataFrame.filter.trueCount === dataFrame.rowCount)
+    return seqCol.categories.filter(Boolean).length;
+
+  const visibleSequences = new Set<string>();
```
drizhina, line 724: too much to be honest. just return dataFrame.filter.truecount and that is it. no need for all this, its too expensive

packages/Bio/src/widgets/sequence-scrolling-widget.ts
```diff
@@ -1,147 +1,740 @@
...
+      hash = hash & hash;
+    }
+    return Math.abs(hash).toString(36);
+  }
+
+  /**
+   * Clear all caches and update invalidation timestamp for force updates
+   */
+  static clearAllCaches(): void {
+    MSAViewportManager.conservationCache = new DG.LruCache<string, number[]>(100);
```
drizhina, line 47: you should clear the previous ones first and then assign them to new variables. I know that garbage collector will do it in some time, but it will still take some time, so better do it like that, easier for GC

drizhina (changes requested):

PR #3382, opened 2025-06-03
Moved composition-table to library, bump bio

rtviii:
Moved `buildCompositionTable` out of `packages/Bio/.../composition-analysis-widget` to `libraries/bio/utils/composition-table`. 
- Note, i left `getCompositionAnalysisWidget` in that same file. Is that right?

Also bumped the library's `package.json` version as @drizhina showed on a call for it to be gh-action propagated to npm. Hopefully got that right as well?


drizhina (approved):
LGTM

PR #3402, opened 2025-06-23
PeptidesSAR: LST WebLogo agg type/column options. First stab.

rtviii:
Hi Davit, 

This is not for submission yet. Could you take a quick looksie if this seems right/you find the changes satisfactory.

The aggregation type/column are added to weblogo opts in LST now, cache invalidation. It seems to work, but i have a bit of a suspicion about the `AggregationColumn` somehow..

Also, unrelated to your assignment but in general for this viewer: i had to uncofuse myself about what `Activity(2)` means (vis-a-vis `Activity). I got that it's the "derived" lg or -lg column, but maybe we should change to something like `Derived lg Activity` just for the user?

### DIFF
--- a/packages/Peptides/src/viewers/logo-summary.ts
+++ b/packages/Peptides/src/viewers/logo-summary.ts
@@ -55,6 +55,9 @@ export const enum LST_PROPERTIES {
   AGGREGATION = 'aggregation',
   ACTIVITY_SCALING = 'activityScaling',
   ACTIVITY = 'activity',
+
+  WEB_LOGO_AGGREGATION_COLUMN = 'webLogoAggregationColumn',
+  WEB_LOGO_AGGREGATION_TYPE = 'webLogoAggregationType',
 }
 
 enum LST_CATEGORIES {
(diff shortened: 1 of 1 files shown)


packages/Peptides/src/viewers/logo-summary.ts
```diff
@@ -122,6 +130,25 @@ export class LogoSummaryTable extends DG.JsViewer implements ILogoSummaryTable {
...
+      DG.AGG.SELECTED_ROWS_COUNT];
+    const aggChoices = Object.values(DG.AGG).filter((agg) => !aggExcludeList.includes(agg));
+    this.webLogoAggregationType = this.string(LST_PROPERTIES.WEB_LOGO_AGGREGATION_TYPE, DG.AGG.TOTAL_COUNT, {
+      category: LST_CATEGORIES.STYLE,
+      choices: aggChoices,
+      description: 'Aggregation method for WebLogo position heights',
+    }) as DG.AggregationType;
+
+
+    this.webLogoAggregationColumn = this.column(LST_PROPERTIES.WEB_LOGO_AGGREGATION_COLUMN, {
```
drizhina, line 143: this will not work. this is a specialty I was talking about, both the property name, and the name of property should end with 'ColumnName'

rtviii: Ah, now this thing comes in play that you mentioned n>10 times. I think this should do it (Updated)

PR #3422, opened 2025-07-14
Peptide-tools -based property calculators and unified UI with annotation for them

rtviii:


rtviii:
This is not final. I still need to populate the actual info context for sources correctly.

Also what do we do about `logS` from aqSol? As in.. do we want to use their default weights through python or go through the motions of porting these models to datagrok's js versions.

packages/Bio/package.json
```diff
@@ -139,7 +139,8 @@
         },
         "Calculate": {
           "Identity...": null,
-          "Similarity...": null
+          "Similarity...": null,
+        "Isoelectric Point (pI)...": null
```
drizhina, line 143: Not needed any more

packages/Bio/src/package.ts
```diff
@@ -1272,3 +1272,159 @@ export function test1(): any {
   _package.logger.debug('Bio:test1() function');
   return {value: 'value1'};
 }
+
+//top-menu: Bio | Calculate | Isoelectric Point (pI)...
+//name: calculatePiMenu
+//description: Calculates isoelectric point (pI) for molecules
+//input: dataframe table [Input data table]
+//input: column molecules {semType: Molecule} [Molecule column]
+export async function calculatePiMenu(table: DG.DataFrame, molecules: DG.Column): Promise<void> {
```
drizhina, line 1281: all of this should be:
1. in chem package
2. in separate file, not in package.ts, this is an entry point
3. should be dynamically created. python functions will return a valid dialog content widget and you shell use that

drizhina (changes requested):

PR #3426, opened 2025-07-22
CPC: Moved to chem, factored into a separate widget file.

rtviii:
- New biochem property scripts are in `packages/Chem/scripts`
- The conda environmet file is at `Chem/environments` now 
- The dynamic function widget is added

Unfortunately we couldn't solve the particular environment that aqSolPred _models_ require to be read with Pavlo (it installs fine on my local machine, but requires an ancient version of scikit to deserialize the `.pkl` models, datagrok env solver times out or something..?). Hence i untagged the `logS` script for now until we port those models to datagrok js idiom as planned.

### DIFF
--- a/packages/Chem/src/package.ts
+++ b/packages/Chem/src/package.ts
@@ -1826,7 +1826,7 @@ export async function callChemDiversitySearch(
 //meta.function_family: biochem-calculator
 //meta.method_info.author: Open Chem Lib Team
 //meta.method_info.year: 2024
-//meta.method_info.package: https://github.com/actelion/openchemlib
+//meta.method_info.github: https://github.com/actelion/openchemlib
 export async function addChemPropertiesColumns(table: DG.DataFrame, molecules: DG.Column,
   MW?: boolean, HBA?: boolean, HBD?: boolean, logP?: boolean, logS?: boolean,
   PSA?: boolean, rotatableBonds?: boolean, stereoCenters?: boolean, moleculeCharge?: boolean,


packages/Chem/src/widgets/biochem-properties-widget.ts
```diff
@@ -0,0 +1,393 @@
...
+    github?: string;
+    citation?: string;
+  }
+
+class HistoryManager {
+  private historyStack: string[] = [];
+  private historyPointer: number = -1;
+  public undoButton: HTMLElement;
+  public redoButton: HTMLElement;
+  public isRestoringState: boolean = false;
```
drizhina, line 33: you got this idea a bit wrong. the history is core feature of the platform, and it manages it by itself. all you needed to do was to provide the .history function to the dialog. and I pointed you to where you should have looked.
eda/src/package.ts line 180 for example
 dialog.history(() => ({editorSettings: editor.getStringInput()}), (x: any) => editor.applyStringInput(x['editorSettings']));
something like this. we do not have the redo undo stuff anywhere

packages/Chem/src/widgets/biochem-properties-widget.ts
```diff
@@ -0,0 +1,393 @@
...
+    ui.empty(moleculesInputDiv);
+    moleculesInputDiv.appendChild(moleculesInput.root);
+  };
+  const tableInput = ui.input.table('Table', {value: table, items: grok.shell.tables, onValueChanged: (t) => {table = t!; onTableChanged();}});
+  onTableChanged();
+  const topInputContainer = ui.divH([tableInput.root, moleculesInputDiv], 'biochem-calc-top-inputs');
+  const inputSection = ui.div([topInputContainer], 'biochem-calc-input-section');
+
+  const functionState = new Map<string, FunctionState>();
+  const historyManager = new HistoryManager(functionState, table);
```
drizhina, line 191: and therefore this needs to go as well

packages/Chem/src/widgets/biochem-properties-widget.ts
```diff
@@ -0,0 +1,393 @@
...
+
+  let debounceTimer: NodeJS.Timeout;
```
drizhina, line 193: ci Typescript will complain on this, safer bet is to give it a type as:
let debounceTimer: NodeJS.Timeout | number | null = null;

packages/Chem/src/widgets/biochem-properties-widget.ts
```diff
@@ -0,0 +1,393 @@
...
+
+  private updateButtonState = (): void => {
+    const undoIcon = this.undoButton.firstChild as HTMLElement;
+    const redoIcon = this.redoButton.firstChild as HTMLElement;
+    if (undoIcon) undoIcon.style.color = this.historyPointer > 0 ? '' : '#c0c0c0';
+    if (redoIcon) redoIcon.style.color = this.historyPointer < this.historyStack.length - 1 ? '' : '#c0c0c0';
+  };
+}
+
+const styles = `
```
drizhina, line 137: add these to css file and import it, there are multiple of such cases in chem and some css as well

packages/Chem/scripts/biochemical-calculators/calc_logD.py
```diff
@@ -0,0 +1,54 @@
+#name: Calculate logD
+#description: Calculates the distribution coefficient (logD) at pH 7.4 using pKa and logP.
+#language: python
+#tags: biochem-calculator
```
drizhina, line 4: also please add this tag to the chem properties calculator function in chem package.ts (OCL) and make sure it works in dialog. it should go first in the dialog

drizhina (changes requested):

packages/Chem/scripts/biochemical-calculators/calc_logD.py
```diff
@@ -0,0 +1,59 @@
+#name: Calculate logD
+#description: Calculates the distribution coefficient (logD) at a specified pH using pKa and logP.
+#language: python
+#meta.function_family: biochem-calculator
+#environment: channels: [conda-forge, defaults], dependencies: [python=3.9, pip, rdkit, {pip: [pichemist]}]
+#input: dataframe table
+#input: string molecules_column_name
```
drizhina, line 7: this input should be of type column, and python will just use string in any case regardless. and lets call variable `molecules` , because column name will be appended to it anyways

packages/Chem/scripts/biochemical-calculators/calc_logP.py
```diff
@@ -0,0 +1,34 @@
+#name: Calculate logP
+#description: Calculates the octanol-water partition coefficient (logP) using RDKit's Crippen method.
+#language: python
+#meta.function_family: biochem-calculator
+#environment: channels: [conda-forge, defaults], dependencies: [python=3.9, pip, rdkit]
```
drizhina, line 5: environment is not needed here, rdkit is preinstalled in default env

packages/Chem/scripts/biochemical-calculators/calc_logS.py
```diff
@@ -0,0 +1,122 @@
+#name: Calculate logS
+#description: Predicts aqueous solubility (logS) using the AqSolPred consensus model.
+#language: python
+#environment: channels: [conda-forge, defaults], dependencies: [python=3.8, scikit-learn=0.23.2, numpy=1.19.5, pandas=1.3.5, rdkit, {pip: [mordred, xgboost]}]
+#input: dataframe table
+#input: string molecules_column_name
```
drizhina, line 6: same for the molecules input, type column and name just molecules. we also want to be able to use these funcs as is from API

packages/Chem/src/widgets/biochem-properties-widget.ts
```diff
@@ -0,0 +1,307 @@
...
+    molecules: DG.Column,
+  ) {
+    const progress = DG.TaskBarProgressIndicator.create('Calculating properties...');
+    let completedCount = 0;
+    try {
+      for (const [funcName, state] of selectedFunctions) {
+        progress.update((completedCount / selectedFunctions.length) * 100, `Calculating ${funcName}...`);
+        state.funcCall.inputs['table'] = table;
+        state.funcCall.inputs['molecules_column_name'] = molecules.name;
+        state.funcCall.inputs['molecules'] = molecules.name;
```
drizhina, line 270: this is a bit unsafe, I suggest you save the name of table and column parameters, and use them, in the same state. also, you can apply a sinply search, instead of finding it using inputs.find, table and column inputs should just be first and second inputs

drizhina (changes requested):
Looks LGTM! Some small comments and then we can merge

PR #3526, opened 2025-11-10
GROK-19146: File input: Ability to select directory

MariaDolotova:


js-api/src/api/grok_api.g.ts
```diff
@@ -607,6 +607,8 @@ export interface IDartApi {
   grok_ColumnsInput_SetAdditionalColumns(input: any, columns: any): any;
   grok_ColumnsInput_SetOnAdditionalColumnsChanged(input: any, fn: any): any;
   grok_ColorInput_SetShowOnlyColorBox(input: any, x: any): any;
+  grok_FileInput_Get_Directory_Files(input: any): any;
```
skalkin: The name does not conform to our standards. If it's a getter, it should be grok_FileInput_Get_DirectoryFiles

PR #3583, opened 2025-12-18
Chem: Migrate tags to role

Aleksashka11:


js-api/src/const.ts
```diff
@@ -403,7 +403,36 @@ export const FUNC_TYPES = {
...
+
+  HIT_TRIAGE_FUNCTION: 'HitTriageFunction',
+  HIT_TRIAGE_DATA_SOURCE: 'HitTriageDataSource',
+  HIT_TRIAGE_SUBMIT_FUNCTION: 'HitTriageSubmitFunction',
+  HIT_DESIGN_FUNCTION: 'HitDesignFunction',
+
+  DIM_RED_PREPROCESS: 'dim-red-preprocessing-function',
+  DIM_RED_POSTPROCESS: 'dim-red-postprocessing-function',
+
+  SC_WEBGPU_POINT_HIT_TEST: 'scWebGPUPointHitTest',
```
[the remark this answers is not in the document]
drizhina: I think I answered this but still:
for webGPU, yes, that is correct.
for Dim Reduction stuff, they are polymorphic, so should be roles

js-api/src/const.ts
```diff
@@ -420,48 +449,48 @@ export const functionRoles: FuncRoleDescription[] = [
...
     header: 'tags',
-    signature: 'app()'
+    signature: 'app(): void'
   },
   {
     role: FUNC_TYPES.PANEL,
     description: 'Context-specific widget that appears on the context panel.',
     header: 'tags',
-    signature: 'panel(x: any): Widget'
+    signature: 'panel(x: any): Widget | graphics'
```
drizhina: x: any can be confusing for llm. x is always either a semanticValue or a primitive

js-api/src/const.ts
```diff
@@ -420,48 +449,48 @@ export const functionRoles: FuncRoleDescription[] = [
     role: FUNC_TYPES.APP,
     description: 'An application that gets shown in the app store.',
     header: 'tags',
-    signature: 'app()'
+    signature: 'app(): void'
```
drizhina: this is tricky. in a perfect world, an app function should return a view.
but we don't live in a perfect world and some views handle adding by themselves and return nothing)) so not sure

js-api/src/const.ts
```diff
@@ -480,20 +509,164 @@ export const functionRoles: FuncRoleDescription[] = [
...
+    role: FUNC_TYPES.HIT_TRIAGE_FUNCTION,
+    description: 'Compute function for Hit Triage campaigns.',
+    header: 'tags',
+    signature: 'hitTriageFunction(table: DataFrame, moleculeCol: Column): DataFrame'
+  },
+  {
+    role: FUNC_TYPES.HIT_TRIAGE_DATA_SOURCE,
+    description: 'Provides a datasource for Hit Triage campaigns. Must return a dataframe containing molecules.',
+    header: 'tags',
+    signature: 'hitTriageDataSource(numberOfMolecules: number]): DataFrame'
```
drizhina: again, numberOfMolecules is an example. arguments can be any length and anything
hitTriageDataSource(...args: any): DataFrame

js-api/src/const.ts
```diff
@@ -480,20 +509,164 @@ export const functionRoles: FuncRoleDescription[] = [
...
+  },
+  {
+    role: FUNC_TYPES.HIT_TRIAGE_SUBMIT_FUNCTION,
+    description: 'Processes or saves the computed dataset.',
+    header: 'tags',
+    signature: 'hitTriageSubmitFunction(df: DG.DataFrame, moleculesCol: string): void'
+  },
+  {
+    role: FUNC_TYPES.HIT_DESIGN_FUNCTION,
```
drizhina: where did that come from :D
hit design reuses the same hit_triage_function role for compute funcs
there is a thing that looks like this which is a HitDesignerFunction
this one is for stuff like reinvent.
also, the signature is similar as above, first arg should be mol and rest args
hitDesignFunction(molecule: string, ...args: any): DataFrame

js-api/src/const.ts
```diff
@@ -480,20 +509,164 @@ export const functionRoles: FuncRoleDescription[] = [
...
+  },
+  {
+    role: FUNC_TYPES.SEARCH_PROVIDER,
+    description: 'Marks a function to be used as a search provider in the global search.',
+    header: 'tags',
+    signature: 'searchProvider(): SearchProvider'
+  },
+  {
+    role: FUNC_TYPES.NOTATION_REFINER,
+    description: '',
```
drizhina: Refines the biological sequence notation based on company specific rules

tools/bin/commands/migrate.ts
```diff
@@ -0,0 +1,110 @@
+import {Project, SyntaxKind, ObjectLiteralExpression, QuoteKind, IndentationText} from 'ts-morph';
```
drizhina: just a suggestion: "migrate" might be a bit of confusing name, as it mostly is referred in context of DBs.
maybe something more descriptive towards tags-roles

js-api/src/const.ts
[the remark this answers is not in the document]
Aleksashka11: I’ve already removed the WebGPU-related things from the roles
I had to do this in a separate branch, as I was continuing work there on other packages, sorry

PR #3716, opened 2026-04-02
[ticket] CLAUDE-3: PowerPack: Fix AddNewColumnDialog crash on undefined codeMirror

PavloPolovyi:
## Summary
- Fixed `TypeError: Cannot read properties of undefined (reading 'state')` in AddNewColumnDialog
- Added null guards for `this.codeMirror` in `onInput` handlers for name/type inputs, `saveInputHistory`, and `loadInputHistory`
- Root cause: dialog is shown before async CodeMirror initialization completes; user interaction with inputs during that window crashes

## Test plan
- [ ] Open Add New Column dialog, quickly type in the Name field before CodeMirror loads — no crash
- [ ] Change column type dropdown before CodeMirror loads — no crash
- [ ] Normal Add New Column workflow still works after CodeMirror is ready

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/packages/PowerPack/CHANGELOG.md
+++ b/packages/PowerPack/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Power Pack changelog
 
+## v.next
+
+* CLAUDE-3: Fixed crash in Add New Column dialog when interacting with inputs before CodeMirror initializes
+
 ## 1.8.0 (2026-03-13)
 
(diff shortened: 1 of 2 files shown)


LaykDimon:
still doesn't resolve the issue + fixed on master, re-tune

PR #3727, opened 2026-04-03
CLAUDE-14: Curves: Guard against undefined fit function in renderCurves

PavloPolovyi:
## Summary
- Added null guard for `fitFunc` in `FitChartCellRenderer.renderCurves` to prevent `TypeError: Cannot read properties of undefined (reading 'y')` when `getSeriesFitFunction` returns undefined for an unregistered fit function name.

## Test plan
- [ ] Open a table with curves column containing an unknown fit function name
- [ ] Verify rendering does not crash and points are still rendered


LaykDimon:
the fix wouldn't wrok anyway + master has already better solutions done

PR #3728, opened 2026-04-03
CLAUDE-27: SequenceTranslator: Handle empty alphabet error in bulk convert

PavloPolovyi:
## Summary
- Wrap `setUnits` call in try-catch in `processConvertBulkButtonClick` to handle empty alphabet gracefully
- Shows a warning message instead of crashing when converted sequences have unrecognizable alphabet

## Test plan
- [ ] Open Oligo Toolkit, switch to Translator tab
- [ ] Convert sequences that produce empty alphabet output
- [ ] Verify warning is shown instead of crash


LaykDimon:
we don't actually fix the issue by this, but rather hiding it - re-tune on it

PR #3729, opened 2026-04-03
CLAUDE-29: Chem: Guard against null TableView in activity cliffs

PavloPolovyi:
## Summary
- Add null guard for `grok.shell.tv` in activity cliffs to prevent crash when no TableView is active

## Test plan
- [ ] Run activity cliffs analysis
- [ ] Navigate away from TableView before analysis completes
- [ ] Verify no crash occurs


LaykDimon:
fixed + the guard doesn't really fix the issue

PR #3730, opened 2026-04-03
CLAUDE-30: Charts Globe: Handle missing columns gracefully

PavloPolovyi:
## Summary
- Use `col()` (returns null) instead of `getCol()` (throws) in Globe viewer `getCoordinates`
- Add null guards for all required columns (latitude, longitude, colorBy, magnitude)
- Clear points and return early when columns are missing

## Test plan
- [ ] Open Globe viewer on a table with lat/lon columns
- [ ] Remove a required column while the viewer is active
- [ ] Verify viewer handles it gracefully without crashing


LaykDimon:
duplicate + already fixed

PR #3732, opened 2026-04-03
[ticket] CLAUDE-14: Curves: Skip series with unknown fit function in renderCurves

PavloPolovyi:
## Summary
- Added null guard for `fitFunc` returned by `getSeriesFitFunction()` in `renderCurves`
- Prevents TypeError when a series references a fit function not registered in the `fitFunctions` registry
- Report: public #2076

## Test plan
- [ ] Open a table with curve data that references an unknown fit function
- [ ] Verify no errors are thrown during rendering


LaykDimon:
the fix wouldn't wrok anyway + master has already better solutions done

PR #3744, opened 2026-04-12
BioStructureViewer: Interactive atom highlighting

talagayev:
Addition of the following feature:

- Highlighting of atoms of the 2D depiction of the molecule in `SMILES` column connected to the 3D visualization and highlighting of docking pose atoms with `Mol*`
- Change of color of the docking pose of compared molecule, when two molecule docking poses are depicted in `Mol*`


.claude/launch.json
```diff
@@ -0,0 +1,17 @@
...
+      "runtimeArgs": ["--yes", "serve", "-l", "4173", "packages/Chem/files/mockups"],
+      "port": 4173
+    },
+    {
+      "name": "docusaurus",
+      "runtimeExecutable": "npm",
+      "runtimeArgs": ["--prefix", "docusaurus-static", "run", "start"],
+      "port": 3000
+    }
+  ]
```
drizhina: you probably should not commit this file

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -43,12 +43,51 @@ import {defaults, molecule3dFileExtensions} from './consts';
...
+  setStructureOverpaint, clearStructureOverpaint,
+} from 'molstar/lib/mol-plugin-state/helpers/structure-overpaint';
+
+/**
+ * Per-row cache of atom selection events. Keyed by rowIdx so that
+ * highlights for different rows (current + hovered ligands) coexist.
+ * Registered at module load so events are captured even before any
+ * Molstar viewer is opened.
+ */
+const _globalSelectionCache = new Map<number, { atoms: number[], mapping3D?: any }>();
```
drizhina: avoid using any type. this shell be removed but in any case, avoid it.

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -43,12 +43,51 @@ import {defaults, molecule3dFileExtensions} from './consts';
...
+
+// Register at module load — no lazy guard needed.
+grok.events.onCustomEvent('chem-interactive-selection-changed')
```
drizhina: this is a bad idea)) 
dataframes might be different, might be closed, might change, there might be two different columns and etc. instead you should use:
DG.LRUCache.
this all should be a static fields on viewer class and singleton-initiated at viewer creation.
the keys should be composite of - DF.id-DF.name-Column.name-rowIDX

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -844,6 +883,58 @@ export class MolstarViewer extends DG.JsViewer implements IBiostructureViewer, I
...
+      // Subscribe at instance level so we can apply highlights in real-time
+      // when the viewer IS alive. Works for BOTH the AutoDock Molstar panel
+      // and the 3D Structure Molstar panel. The module-level listener
+      // (registered at module load) handles caching for replay.
+      this.viewSubs.push(grok.events.onCustomEvent(
+        'chem-interactive-selection-changed').subscribe((args: any) => {
+        try {
+          if (args?.atoms?.length >= 0) {
+            // eslint-disable-next-line no-console
+            console.log('[molstar-picker] live highlight',
```
drizhina: use _package.logger.debug for such prints

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -1165,6 +1270,430 @@ export class MolstarViewer extends DG.JsViewer implements IBiostructureViewer, I
...
+    }
+  }
+
+  /** Reads the highlighted atom indices for a given row from the SMILES
+   *  column's temp substruct providers. Returns empty array if none. */
+  private _getHighlightedAtomsFromColumn(rowIdx: number): number[] {
+    const df = this.dataFrame;
+    if (!df) return [];
+    // Find the molecule column (SMILES).
+    const molCol = df.columns.toList().find(
```
drizhina: this looks like super ambiguous to me tbh. there should be 2 way tags on both columns that say that they are connected. there might be multiple molecule columns in table.

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -1165,6 +1270,430 @@ export class MolstarViewer extends DG.JsViewer implements IBiostructureViewer, I
...
+        .filter((idx) => idx >= 0)
+        .map((idx) => {
+          // If we have actual PDB serial numbers, use them.
+          // Otherwise fall back to 0-based index + 1.
+          if (pdbSerials && idx >= 0 && idx < pdbSerials.length)
+            return pdbSerials[idx];
+          return idx + 1;
+        });
+      // eslint-disable-next-line no-console
+      console.log('[molstar-picker] _computeSerials: using mapping3D (method:',
```
drizhina: here as well. replace console logs with package logger debug

packages/Chem/src/rendering/rdkit-cell-renderer.ts
```diff
@@ -69,6 +85,36 @@ export function _addColorsToBondsAndAtoms(mainSubstr: ISubstruct, color?: string
...
+    // chem-meta change but the published `@datagrok-libraries/chem-meta`
+    // package on npm is still the older 1.x without them. CI fetches the
+    // npm version, so the type check fails on a fresh `npm install`.
+    // Casts erase the type check; runtime is unaffected because the
+    // properties are optional and the runtime JS doesn't care about
+    // TypeScript types. Once chem-meta is republished with the new
+    // fields, these casts can be removed.
+    if (tempSubstr) {
+      const t = tempSubstr as any;
+      const m = mainSubstr as any;
```
drizhina: are these as any casts needed? I believe you did update the type of Isubstrict from library but don't see in files

packages/Chem/src/rendering/rdkit-cell-renderer.ts
```diff
@@ -125,6 +208,27 @@ M  END
...
+    const df = col.dataFrame;
+    if (!df) return false;
+    if (this._pickerActiveCache?.dfId === df.id)
+      return this._pickerActiveCache.active;
+    const cols = df.columns.toList();
+    const active =
+      cols.some((c) => c.semType === DG.SEMTYPE.MOLECULE3D) ||
+      cols.some((c) => c.semType === (DG.SEMTYPE as any).MACROMOLECULE && (c.meta as any)?.units === 'helm');
+    this._pickerActiveCache = {dfId: df.id, active};
+    return active;
```
drizhina: similarly here. this is too vague. it should not be autodetected but should be based on tags. also same argument, there might be multiple molecule columns

packages/Chem/src/rendering/rdkit-cell-renderer.ts
```diff
@@ -457,6 +561,23 @@ M  END
...
+    // current cell and the context panel (e.g. AutoDock) stays open.
+    if (!this._documentListenerAttached) {
+      this._documentListenerAttached = true;
+      document.addEventListener('mousemove', (e: MouseEvent) => {
+        this._onDocumentMouseMove(e);
+      });
+      document.addEventListener('mousedown', (e: MouseEvent) => {
+        this._onDocumentMouseDown(e);
+      }, true);
+      document.addEventListener('keydown', (e: KeyboardEvent) => {
```
drizhina: the renderer class provides overridable methods for each of the events above. look into other renderers. no need to use document events

drizhina: for example look into Curves plugin and fit-cell-renderer file to see these overrides

packages/Chem/src/rendering/rdkit-cell-renderer.ts
```diff
@@ -605,6 +745,684 @@ M  END
...
+   * the persistent selection. The selection accumulates until cleared
+   * with Escape.
+   */
+  private _onDocumentMouseMove(e: MouseEvent): void {
+    // Only activate the picker when the current cell is a Molecule3D
+    // (docking pose) or HELM column. If the user clicked on SMILES,
+    // binding energy, or any other column, hovering does nothing.
+    const df = grok.shell.tv?.grid?.dataFrame;
+    if (df) {
+      const curCol = df.currentCol;
```
drizhina: same here, connection between columns should be tags based

packages/Chem/src/rendering/rdkit-cell-renderer.ts
```diff
@@ -605,6 +745,684 @@ M  END
...
+      // Replace the getSubstruct callback with updated atoms.
+      prov.getSubstruct = (ridx) => {
+        if (!this._isPickerActive(col)) return undefined;
+        if (ridx !== rowIdx) return undefined;
+        return {atoms: atomsArr, bonds: bondsArr, highlightAtomColors, highlightBondColors};
+      };
+    }
+
+    // Clear render cache + invalidate (same pattern as Escape).
+    this.rendersCache = new DG.LruCache<String, ImageData>();
```
drizhina: clear the previous one, it will help GC to collect it easier

packages/Chem/src/rendering/rdkit-cell-renderer.ts
```diff
@@ -605,6 +745,684 @@ M  END
...
+    // viewer) about the updated selection so they can mirror the highlights.
+    // Include a 3D atom-index mapping if a MOLECULE3D column exists in the
+    // same dataframe (e.g. AutoDock poses). The mapping is computed via
+    // RDKit substructure match between the 2D SMILES and the 3D pose, so
+    // BiostructureViewer doesn't need to import RDKit or the mapper.
+    const eventArgs: any = {column: col, rowIdx, atoms: atomsArr};
+    try {
+      const df = col.dataFrame;
+      if (df) {
+        const mol3DCol = df.columns.toList().find(
```
drizhina: same here. pretty much everywhere the comment applies that connection should be explicit

drizhina (changes requested):

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
talagayev: remove any tyoes in const

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
talagayev: adjusted to use `.logger.debug`.

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
talagayev: Adjusted the detection :)

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -1165,6 +1313,432 @@ export class MolstarViewer extends DG.JsViewer implements IBiostructureViewer, I
...
+          // from it by walking up the state tree.
+          if (h.dataRef) {
+            let found = false;
+            try {
+              // Find the state cell for this structure and walk parents.
+              for (const [, cell] of plugin.state.data.cells) {
+                if (cell.obj?.data === structureData) {
+                  let cur: any = cell;
+                  for (let d = 0; cur && d < 6; d++) {
+                    if (cur.transform?.ref === h.dataRef) { found = true; break; }
```
Aleksashka11: This isn’t aligned with our preferred coding style
In most cases, each statement should be placed on its own line, it will significantly improve readability

packages/BiostructureViewer/src/tests/atom-highlight-tests.ts
```diff
@@ -0,0 +1,108 @@
...
+      binary: true, ext: 'pdb',
+      data: await _package.files.readAsBytes('samples/1bdq-wo-ligands.pdb'),
+    };
+    const sdfCnt = await _package.files.readAsBytes('samples/1bdq-obs-pred.sdf');
+    const df: DG.DataFrame = (await grok.functions.call('Chem:importSdf', {bytes: sdfCnt}))[0];
+    const ligandCol = df.getCol('molecule');
+
+    const view = grok.shell.addTableView(df);
+    await awaitGrid(view.grid);
+
```
Aleksashka11: I’ve noticed that several tests contain duplicated setup code, for example:

```
const pdbData: BiostructureData = {
  binary: true,
  ext: 'pdb',
  data: await _package.files.readAsBytes('samples/1bdq-wo-ligands.pdb'),
};
const sdfCnt = await _package.files.readAsBytes('samples/1bdq-obs-pred.sdf');
...
```

We can move this shared setup into the before and reuse it across tests instead of duplicating it
This will also help reduce test runtime, which is something we should care about too

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -43,12 +43,96 @@ import {defaults, molecule3dFileExtensions} from './consts';
...
+/** Atom-index mapping produced by Chem's mapAtomIndices2Dto3D.
+ *  Mirrors the AtomIndexMapping interface from atom-index-mapper.ts
+ *  (defined here to avoid a cross-package import). */
+interface AtomMapping3D {
+  mapping: number[];
+  method: string;
+  mappedCount: number;
+  pdbSerials?: number[];
+}
+
```
Aleksashka11: Ideally, where feasible, I would move everything that isn’t directly related to the Molstar viewer itself into separate components or utility modules and reuse it from there

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -43,12 +43,96 @@ import {defaults, molecule3dFileExtensions} from './consts';
...
+// Register at module load — no lazy guard needed.
+grok.events.onCustomEvent('chem-interactive-selection-changed')
+  .subscribe((_args: unknown) => {
+    const args = _args as ChemSelectionEventArgs;
+    const rowIdx = args?.rowIdx ?? -1;
+    const atoms = args?.atoms ?? [];
+    const isPersistent = args?.persistent !== false;
+    const col = args?.column as DG.Column | undefined;
```
Aleksashka11: Personally, I prefer avoiding stuff like that and using destructuring whenever feasible, e.g.
`const { rowIdx = -1, atoms = [], column } = args ?? {};`

But this is up to you, this is mostly a matter of style

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -844,6 +928,56 @@ export class MolstarViewer extends DG.JsViewer implements IBiostructureViewer, I
       if (!(this.dataEff.ext in molecule3dFileExtensions))
         throw new Error(`Unsupported file extension '${this.dataEff.ext}'.`);
 
+      // Subscribe at instance level so we can apply highlights in real-time
+      // when the viewer IS alive. Works for BOTH the AutoDock Molstar panel
+      // and the 3D Structure Molstar panel. The module-level listener
+      // (registered at module load) handles caching for replay.
+      this.viewSubs.push(grok.events.onCustomEvent(
+        'chem-interactive-selection-changed').subscribe((_args: unknown) => {
```
Aleksashka11: I see there’s already a constant defined for this — let’s use it :)

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -1165,6 +1313,432 @@ export class MolstarViewer extends DG.JsViewer implements IBiostructureViewer, I
...
+   *  unconditionally when 2+ poses are loaded. */
+  private async _applyBaseColors(): Promise<void> {
+    const plugin = this.viewer?.plugin;
+    if (!plugin) return;
+
+    const loadedLigands: { rowIdx: number, structureRefs: string[] | null }[] = [];
+    const seenRows = new Set<number>();
+    const lig = this.ligands;
+    if (lig) {
+      if (lig.current?.rowIdx != null && lig.current.rowIdx >= 0 && !seenRows.has(lig.current.rowIdx)) {
```
Aleksashka11: I think the cleaner approach is to assign it to a variable instead of repeating lig.current.rowIdx 3 times:
`const rowIdx = lig.current?.rowIdx;`

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -1165,6 +1313,432 @@ export class MolstarViewer extends DG.JsViewer implements IBiostructureViewer, I
...
+          if (sel.rowIdx >= 0 && !seenRows.has(sel.rowIdx)) {
+            loadedLigands.push(sel);
+            seenRows.add(sel.rowIdx);
+          }
+        }
+      }
+    }
+
+    this.logger.debug('[molstar-picker] loaded ligands: ' +
+      JSON.stringify(loadedLigands.map((l) => ({row: l.rowIdx, hasRefs: !!l.structureRefs}))));
```
Aleksashka11: I think we should be able to do something like this:

```
this.logger.debug('[molstar-picker] loaded ligands:', () =>
  loadedLigands.map(l => ({
    row: l.rowIdx,
    hasRefs: !!l.structureRefs,
  }))
);

```
I don’t think the JSON.stringify output is very readable in logs anyway

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
Aleksashka11: There are quite a few similar cases across the codebase. I didn’t comment on each one individually, so please take another pass through the code

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -1165,6 +1313,432 @@ export class MolstarViewer extends DG.JsViewer implements IBiostructureViewer, I
...
+      const serials = atomIndices
+        .map((i) => mapping3D.mapping[i])
+        .filter((idx) => idx >= 0)
+        .map((idx) => {
+          // If we have actual PDB serial numbers, use them.
+          // Otherwise fall back to 0-based index + 1.
+          if (pdbSerials && idx >= 0 && idx < pdbSerials.length)
+            return pdbSerials[idx];
+          return idx + 1;
+        });
```
Aleksashka11: Oh, wouldn’t it be better to do this in a single for...of pass?

packages/Chem/src/rendering/rdkit-cell-renderer.ts
```diff
@@ -605,6 +745,684 @@ M  END
...
+  private _onDocumentMouseMove(e: MouseEvent): void {
+    // Only activate the picker when the current cell is a Molecule3D
+    // (docking pose) or HELM column. If the user clicked on SMILES,
+    // binding energy, or any other column, hovering does nothing.
+    const df = grok.shell.tv?.grid?.dataFrame;
+    if (df) {
+      const curCol = df.currentCol;
+      const is3D = curCol?.semType === DG.SEMTYPE.MOLECULE3D;
+      const isHelm = curCol?.semType === (DG.SEMTYPE as any).MACROMOLECULE &&
+        (curCol?.meta as any)?.units === 'helm';
```
Aleksashka11: Why is there an as any cast here? This should work directly as curCol.meta.units
Same applies here: (DG.SEMTYPE as any) — this looks suspicious

packages/Chem/src/rendering/rdkit-cell-renderer.ts
```diff
@@ -605,6 +745,684 @@ M  END
...
+      const colName = col.name;
+      const rowIdx = hit.gridCell.tableRowIndex!;
+
+      // Same atom — skip.
+      if (this._lastHoveredAtom &&
+          this._lastHoveredAtom.col === colName &&
+          this._lastHoveredAtom.rowIdx === rowIdx &&
+          this._lastHoveredAtom.atomIdx === hit.nearest) return;
+      this._lastHoveredAtom = {col: colName, rowIdx, atomIdx: hit.nearest};
+
```
Aleksashka11: There are also repeated pieces of logic, e.g.:

```
this._lastHoveredAtom &&
  this._lastHoveredAtom.col === colName &&
  this._lastHoveredAtom.rowIdx === rowIdx &&
  this._lastHoveredAtom.atomIdx === hit?.nearest
```

And branching can definitely be refactored to improve readability

packages/BiostructureViewer/src/tests/atom-highlight-tests.ts
Aleksashka11: Applies to Chem tests too

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
Aleksashka11: Also, I strongly suggest moving the highlighting component out of molstar-viewer.ts, if possible
This file was already quite large, and it’s only getting bigger now, which makes it harder to maintain

libraries/bio/src/viewers/molecule3d.ts
```diff
@@ -7,6 +7,16 @@ export type Molecule3DData = {
...
 
+/** Persistent column tag used to link a SMILES (Molecule) column to a
+ *  specific Molecule3D column for the Chem atom-picker bridge. The tag
+ *  is set on the SMILES column; its value is the linked Molecule3D column
+ *  name. Canonical constant owned by Chem's `constants.ts`; re-exported
+ *  from the bio library so BiostructureViewer, Docking and other packages
+ *  that already depend on `@datagrok-libraries/bio` can reference it
+ *  without introducing a hard dependency on `@datagrok/chem`. Keep the
+ *  string value in sync with Chem's `CHEM_ATOM_PICKER_LINKED_COL`. */
+export const CHEM_ATOM_PICKER_LINKED_COL = '%chem-atom-picker-linked-col';
```
drizhina: should start with .% not just %

packages/BiostructureViewer/src/utils/mol3d-link.ts
```diff
@@ -0,0 +1,102 @@
...
+export const CHEM_ATOM_PICKER_LINKED_COL = '%chem-atom-picker-linked-col';
+
+/** Mirrors CHEM_ATOM_PICKER_LINKED_SMILES_COL from Chem's constants.ts. */
+export const CHEM_ATOM_PICKER_LINKED_SMILES_COL = '%chem-atom-picker-linked-smiles-col';
+
+/** Mirrors CHEM_INTERACTIVE_SELECTION_EVENT from Chem's constants.ts. */
+const CHEM_SELECTION_EVENT = 'chem-interactive-selection-changed';
+
+/** Mirrors ChemTemps.SUBSTRUCT_PROVIDERS from @datagrok-libraries/chem-meta. */
+const SUBSTRUCT_PROVIDERS_KEY = 'substruct-providers';
```
drizhina: why don't you import?
everything above as well. add those in the chem-meta library. biostructureviewer package already has that library installed. the comment on .% is valid

packages/BiostructureViewer/src/utils/mol3d-link.ts
```diff
@@ -0,0 +1,102 @@
...
+    (col) => col.tags[CHEM_ATOM_PICKER_LINKED_COL] === mol3DColName);
+  return c?.name ?? null;
+}
+
+/** Removes all picker links that point at `mol3DColName` and clears residual highlights. */
+export function clearLinksToMol3D(
+  df: DG.DataFrame, mol3DColName: string): void {
+  for (const c of df.columns.toList()) {
+    if (c.tags[CHEM_ATOM_PICKER_LINKED_COL] === mol3DColName) {
+      delete c.tags[CHEM_ATOM_PICKER_LINKED_COL];
```
drizhina: doesn't really work this way with tags. another specific to how DG dart proxies work. it might work but not sure. we should do `c.tags.delete`
tags is a dart object proxy, so calling delete on it is a bit wrong. please check other places as well

packages/BiostructureViewer/src/utils/mol3d-link.ts
```diff
@@ -0,0 +1,102 @@
...
+      clearAtomPickerHighlights(c);
+    }
+  }
+  const mol3DCol = df.col(mol3DColName);
+  if (mol3DCol?.tags[CHEM_ATOM_PICKER_LINKED_SMILES_COL])
+    delete mol3DCol.tags[CHEM_ATOM_PICKER_LINKED_SMILES_COL];
```
drizhina, line 43: same

packages/BiostructureViewer/src/utils/mol3d-link.ts
```diff
@@ -0,0 +1,102 @@
...
+/** Writes or clears the reciprocal-tag link between `smilesCol` and `mol3DColName`.
+ *  Pass null to unlink (clears both forward and reverse tags + highlights). */
+export function setSmilesColLink(
+  smilesCol: DG.Column, mol3DColName: string | null): void {
+  const df = smilesCol.dataFrame;
+  const prevMol3DName = smilesCol.tags[CHEM_ATOM_PICKER_LINKED_COL];
+  const prevMol3DCol = prevMol3DName && df ? df.col(prevMol3DName) : null;
+  if (prevMol3DCol && prevMol3DCol.name !== mol3DColName &&
+      prevMol3DCol.tags[CHEM_ATOM_PICKER_LINKED_SMILES_COL])
+    delete prevMol3DCol.tags[CHEM_ATOM_PICKER_LINKED_SMILES_COL];
```
drizhina, line 55: same, and one bellow

packages/BiostructureViewer/src/utils/mol3d-link.ts
```diff
@@ -0,0 +1,102 @@
...
+}
+
+/** Strips picker providers, fires clear-all event, and invalidates the grid. */
+export function clearAtomPickerHighlights(smilesCol: DG.Column): void {
+  try {
+    const providers = (smilesCol.temp?.[SUBSTRUCT_PROVIDERS_KEY] ?? []) as
+      Array<{__atomPicker?: boolean}>;
+    if (providers.length > 0) {
+      smilesCol.temp[SUBSTRUCT_PROVIDERS_KEY] = providers.filter(
+        (p) => !p.__atomPicker);
```
drizhina, line 77: a bit weird formatting, please move to upper line

packages/BiostructureViewer/src/utils/mol3d-link.ts
```diff
@@ -0,0 +1,102 @@
...
+    }
+    grok.events.fireCustomEvent(CHEM_SELECTION_EVENT, {
```
drizhina: for the events, might be good idea to create a set of helper functions in chem-meta library (to link it locally, replace chem-meta version number with ../../chem-meta and do npm i) that handle either firing or subscribing to an event. the benefit will be strong types

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-highlight-controller.ts
```diff
@@ -0,0 +1,303 @@
...
+   *  structure rebuilds (overpaint applied before a rebuild is discarded). */
+  replayHighlightIfCached(): void {
+    this.host.viewSyncer.sync('replayHighlight', async () => {
+      try {
+        const plugin = this.host.getPlugin();
+        const comps = this._getAllComponents();
+        if (comps && plugin)
+          await clearStructureOverpaint(plugin, comps);
+      } catch {
+        /* best-effort */
```
drizhina, line 93: maybe log? :D

packages/Chem/src/rendering/rdkit-cell-renderer.ts
```diff
@@ -123,6 +175,57 @@ M  END
...
+      _getCellAtomPositions: this._getCellAtomPositions.bind(this),
+      _clearRendersCache: this._clearRendersCache.bind(this),
+    });
+  }
+
+  /** Clears the render cache so the 2D cell fully redraws.
+   *  DG.LruCache has no public `.clear()`, so we zero the backing arrays
+   *  via a narrow cast to unblock GC of canvas-backed ImageData entries. */
+  private _clearRendersCache(): void {
+    const old = this.rendersCache as unknown as {
```
drizhina: Uf, this is not the way to do it definitely :D :D 
Just creating a new LruCache and assiging it to instance varaiable is enaugh :D 

Mainly, this method should not be used at all. not sure why you need it (clearRendersCache) or cleaning the LRU. please remove it

packages/Chem/src/rendering/atom-picker-controller.ts
```diff
@@ -0,0 +1,539 @@
...
+    return typeof fromTemp === 'string' ? fromTemp : null;
+  }
+
+  // -- Render-time hooks ----------------------------------------------------
+
+  /** Subscribes once to the 3D→2D bridge fired by BiostructureViewer's Molstar viewer. */
+  ensureGlobalListeners(): void {
+    if (this._globalListenersAttached) return;
+    this._globalListenersAttached = true;
+    grok.events.onCustomEvent(CHEM_MOL3D_HOVER_EVENT).subscribe(
```
drizhina: you should not leave subscriptions like that, they should be lifecycle-managed. on another thought, here it is fine.

packages/BiostructureViewer/src/viewers/molstar-viewer/molstar-viewer.ts
```diff
@@ -327,6 +345,43 @@ export class MolstarViewer extends DG.JsViewer implements IBiostructureViewer, I
...
+  public static selectionCache: DG.LruCache<string, SelectionCacheEntry> =
+    new DG.LruCache<string, SelectionCacheEntry>();
+
+  private static _selectionSubscribed = false;
+
+  /** Subscribes once to CHEM_SELECTION_EVENT per process lifetime. */
+  private static _initSelectionCache(): void {
+    if (MolstarViewer._selectionSubscribed) return;
+    MolstarViewer._selectionSubscribed = true;
+    grok.events.onCustomEvent(CHEM_ATOM_SELECTION_EVENT)
```
drizhina: this.subs.push

PR #3755, opened 2026-04-15
Docking: Add hover comparison for docking pose energy properties

talagayev:
When hovering over a different docking pose, the context panel now shows a side-by-side comparison of energy values

Changes made:

- Implementation of comparison of docking pose `binding energy`, `electrostatic `etc. between two molecules

Here is the current viusalization:
Screenshot 2026-04-15 134031

### DIFF
--- a/packages/Docking/src/package.ts
+++ b/packages/Docking/src/package.ts
@@ -12,7 +12,7 @@ import {_runAutodock, AutoDockService, _runAutodock2, ensureNoDockingError} from
 import {TARGET_PATH, BINDING_ENERGY_COL, POSE_COL, BINDING_ENERGY_COL_UNUSED, POSE_COL_UNUSED, ERROR_COL_NAME, ERROR_MESSAGE, AUTODOCK_PROPERTY_DESCRIPTIONS} from './utils/constants';
 import { _demoDocking } from './demo/demo';
 import { DockingViewApp } from './demo/docking-app';
-import { addColorCoding, buildComparisonTable, formatColumns, getFromPdb, getFromPdbs, getReceptorData, processAutodockResults, prop } from './utils/utils';
+import { addColorCoding, buildComparisonTable, formatColumns, getRemarksFromPdb, getRemarksFromPdbs, getReceptorData, processAutodockResults, prop } from './utils/utils';
 export * from './package.g';
 export const _package = new DG.Package();
 
@@ -188,7 +188,7 @@ export class PackageFunctions{
     if (!addedToPdb)
       return new DG.Widget(ui.divText('Docking has not been run'));
 
-    const autodockResults: DG.DataFrame = getFromPdbs(molecule);
+    const autodockResults: DG.DataFrame = getRemarksFromPdbs(molecule);
     const widget = new DG.Widget(ui.div([]));
 
     if (table)
@@ -219,7 +219,7 @@ export class PackageFunctions{
 
     const currentRowIdx = molecule.cell.rowIndex;
     const ligandCol = molecule.cell.column;
-    const currentValues = getFromPdb(ligandCol.get(currentRowIdx));
+    const currentValues = getRemarksFromPdb(ligandCol.get(currentRowIdx));
 
     let debounceTimer: ReturnType<typeof setTimeout> | null = null;
     const sub = currentTable.onMouseOverRowChanged.subscribe(() => {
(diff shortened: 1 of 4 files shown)


talagayev:
We can see if it is currently too crowded, if the difference values should stay or should be remove and regarding the alignment if there are good options for that 🤔

packages/Docking/src/utils/utils.ts
```diff
@@ -7,26 +7,28 @@ import { BiostructureData } from '@datagrok-libraries/bio/src/pdb/types';
 
 import { BINDING_ENERGY_COL, POSE_COL, setAffinity, setPose, TARGET_PATH } from './constants';
 
+export function getFromPdb(pdbString: string): { [name: string]: number } {
```
drizhina: function name is a bit ambiguous. maybe getRemarksFromPdb?

packages/Docking/src/utils/utils.ts
```diff
@@ -7,26 +7,28 @@ import { BiostructureData } from '@datagrok-libraries/bio/src/pdb/types';
...
+  const result: { [name: string]: number } = {};
+  const remarkRegex = /REMARK\s+\d+\s+([\w\s\(\)-]+)\.\s+([-\d.]+)/g;
+  let match;
+  while ((match = remarkRegex.exec(pdbString)) !== null)
+    result[match[1].trim()] = parseFloat(match[2].trim());
+  return result;
+}
+
 export function getFromPdbs(pdb: DG.SemanticValue): DG.DataFrame {
```
drizhina: same here. I know it was there before but lets rename to keep consistant

packages/Docking/src/package.ts
```diff
@@ -205,15 +205,40 @@ export class PackageFunctions{
...
+    };
+    result.appendChild(buildPropertyMap());
     widget.root.append(result);
 
+    const currentRowIdx = molecule.cell.rowIndex;
+    const ligandCol = molecule.cell.column;
+    const currentValues = getFromPdb(ligandCol.get(currentRowIdx));
+
+    let debounceTimer: ReturnType<typeof setTimeout> | null = null;
+    const sub = currentTable.onMouseOverRowChanged.subscribe(() => {
```
drizhina: use the DG.debounce method, that accepts the event
looks like DG.debounce(currentTable.onMouseOverRowChanged, 50).subscribe
instead of timers

packages/Docking/src/package.ts
```diff
@@ -205,15 +205,40 @@ export class PackageFunctions{
...
+      if (debounceTimer) clearTimeout(debounceTimer);
+      debounceTimer = setTimeout(() => {
+        const hovIdx = currentTable.mouseOverRowIdx;
+        result.innerHTML = '';
```
drizhina: here, use ui.empty(result).
this basically also takes care of any inner subscribtions and their lifecycles. its less relevant here, but generally a better way

drizhina (changes requested):

packages/Docking/src/utils/utils.ts
```diff
@@ -86,7 +88,43 @@ export function prop(molecule: DG.SemanticValue, propertyCol: DG.Column, host: H
     .css('margin-right', '5px');
   
   const idx = molecule.cell.rowIndex;
-  return ui.divH([addColumnIcon, propertyCol.get(idx)], {style: {'position': 'relative'}});
+  const val: number = propertyCol.get(idx);
+  const numStyle = {style: {width: '42px', textAlign: 'right', flexShrink: '0', flexGrow: '0'}};
```
Aleksashka11, line 92: Where possible, I prefer to keep styles in a separate CSS file (especially when there’s more than one involved)
I see that Docking currently doesn’t include one, so I’ll go ahead and add it myself and move this styling
So no action required on this one for you, it will be on me :)

packages/Docking/src/utils/utils.ts
```diff
@@ -86,7 +88,43 @@ export function prop(molecule: DG.SemanticValue, propertyCol: DG.Column, host: H
...
+    const sign = delta >= 0 ? '+' : '';
+    const color = delta < 0 ? 'var(--green-2)' : delta > 0 ? 'var(--red-2)' : '';
+    const row = ui.div([
+      ui.divText(cur.toFixed(2)),
+      ui.divText('vs'),
+      ui.divText(hov.toFixed(2)),
+      ui.divText(`\u0394${sign}${delta.toFixed(2)}`),
+    ]);
+    row.style.cssText =
+      'display:grid;grid-template-columns:42px 16px 42px auto;align-items:baseline;column-gap:12px;';
```
Aleksashka11, line 119: Please avoid doing it this way

packages/Docking/src/utils/utils.ts
talagayev: Ah great thank you :) Should I help somehow or modify something or just keep that PR and branch as it is currently?

packages/Docking/src/utils/utils.ts
talagayev: This is connected to the previous comment so that that this part connected to CSS should also be in the separate CSS file? :)

PR #3761, opened 2026-04-21
BiostructureViewer: Protein side chain visualization

talagayev:
This addition allows to visualize the protein side chains inside the binding site.
The small molecule is detected either through reading the `PDB` file or using information from `Docking` cells and upon pressing a button inside `MOL*` opens a window that allows the selection of the radius around the ligand that the protein side chains are going to be displayed.

Changes made:
- Addition of protein binding site side chain visualization in `Mol*`

The following GIF displays an example of the usage of this feature:

protein_side_chain_vis3


packages/BiostructureViewer/src/tests/binding-site-tests.ts
```diff
@@ -0,0 +1,152 @@
...
+    expect((viewer as any).bindingSiteRefs.length, 0, 'bindingSiteRefs empty before any toggle');
+  }, {timeout: 20000});
+
+  test('pathA-het-in-pdb', async () => {
+    // 1bdq.pdb contains HET ligand atoms; the viewer should build a binding
+    // site (component + representation) when toggled on.
+    const {viewer} = await createPathAViewerFromSample('samples/1bdq.pdb');
+
+    await toggleBindingSite(viewer, true);
+    await awaitCheck(() => (viewer as any).bindingSiteRefs.length >= 2,
```
Aleksashka11: Why are we casting viewer to any?

packages/BiostructureViewer/src/tests/binding-site-tests.ts
```diff
@@ -0,0 +1,152 @@
...
+      'Path A binding site refs not created', 10000);
+  }, {timeout: 45000});
+
+  test('pathA-apo-no-het', async () => {
+    // 1bdq-wo-ligands.pdb has ligand HETs stripped. Toggle should be a noop.
+    const {viewer} = await createPathAViewerFromSample('samples/1bdq-wo-ligands.pdb');
+
+    expect((viewer as any)._isLigandAvailable(), false,
+      'Apo structure should not report a ligand');
+
```
Aleksashka11, line 83: I think this should be a single line. It clearly doesn’t exceed the 120-character limit, and it remains readable on screen

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+      if (resIdx !== prevResIdx) {
+        if (prevResIdx !== -1) commit();
+        prevResIdx = resIdx;
+        curCompId = SP.atom.label_comp_id(loc);
+        curChainId = SP.chain.label_asym_id(loc);
+        curSeqId = SP.residue.label_seq_id(loc);
+        curCount = 0;
+      }
+      const element = SP.atom.type_symbol(loc);
+      if (element !== 'H' && element !== 'D') curCount++;
```
Aleksashka11: Ideally, we would like to see curCount++ on the next line — it’s more readable that way

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  const positions: number[] = [];
+  for (const unit of ligandStructure.units) {
+    const {elements, conformation} = unit;
+    for (let i = 0, n = elements.length; i < n; i++) {
+      const eIdx = elements[i];
+      positions.push(conformation.x(eIdx), conformation.y(eIdx), conformation.z(eIdx));
+    }
+  }
+  if (positions.length === 0) return null;
+  return new Float32Array(positions);
```
Aleksashka11: Do we really need the intermediate number[] allocation? I think we can write directly into a Float32Array
Also, if we can determine the final size in advance, it’s better to allocate the exact buffer upfront rather than growing an array dynamically
And in general, I prefer index-based assignment over push — it’s typically faster

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+ * Positions the overlay inside its parent (ideally `.msp-viewport`) so its
+ * top edge sits just below the Mol* icon strip when present, or at the
+ * parent's top-right corner otherwise. The overlay's `right` is kept at
+ * 10 px so it aligns with the strip's right edge (strip is also at right:10).
+ * @param {HTMLElement} wrapper the overlay element; its `top` / `right` are updated in place.
+ */
+function positionOverlayRelativeToStrip(wrapper: HTMLElement): void {
+  const container = wrapper.parentElement;
+  if (!container) return;
+  wrapper.style.right = '10px';
```
Aleksashka11: Where possible, styles should be moved to the css file

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+/**
+ * Creates the binding-site overlay DOM element.
+ * The element is positioned absolute to visually extend the .msp-viewport-controls
+ * strip downward. The caller appends it to viewer.root, then should call
+ * `(wrapper as any)._bsvReposition()` once the Mol* strip is in the DOM.
+ * @param {BindingSiteOverlayHandlers} handlers state getters/setters and predicates owned by the viewer.
+ * @return {HTMLElement} the overlay root element with expando `_bsvReposition` / `_bsvRefreshUI` / `_bsvCleanup`.
+ */
+export function createBindingSiteOverlay(handlers: BindingSiteOverlayHandlers): HTMLElement {
+  const wrapper = document.createElement('div');
```
Aleksashka11: Is there a specific reason we’re not using native Datagrok components here (e.g. ui.div())?

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  wrapper.className = 'bsv-bs-overlay';
+  wrapper.style.cssText = [
```
Aleksashka11: Is it really that restrictive? Are we unable to override styles there, even with !important?
I’m just trying to understand whether this can be moved to the cdd file or if there’s some limitation preventing it

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  iconBg.style.cssText = [
+    'position:absolute',
+    'inset:0',
+    'background:#eeece7',
+    'opacity:0.5',
+    'pointer-events:none',
+  ].join(';');
+  iconHolder.appendChild(iconBg);
+
+  const iconBtn = document.createElement('button');
```
Aleksashka11: We already have native components for both buttons and icons — let’s use those instead

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  const popHeader = document.createElement('div');
+  popHeader.style.cssText = [
+    'padding:3px 10px', 'font-size:14px', 'font-weight:bold', 'color:#332b1f',
+    'background:#eeece7', 'display:flex', 'align-items:center', 'gap:6px', 'line-height:24px',
+  ].join(';');
+  popHeader.innerHTML =
+    BULLSEYE_SVG
+      .replace('width="16" height="16"', 'width="13" height="13"')
+      .replace(/currentColor/g, '#332b1f') +
+    '<span>Binding Site</span>';
```
Aleksashka11: That all looks pretty hacky :))

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  const showRow = document.createElement('label');
+  showRow.style.cssText = ROW_STYLE + ';cursor:pointer';
+  const showLbl = document.createElement('span');
+  showLbl.style.cssText = LABEL_STYLE;
+  showLbl.textContent = 'Show side chains';
+  const showCtrl = document.createElement('span');
+  // Checkbox is much narrower than the slider; center it so it sits over the
+  // same x-range the slider's thumb traverses on the Radius row below.
+  showCtrl.style.cssText = CONTROL_STYLE + ';justify-content:center';
+  const showChk = document.createElement('input');
```
Aleksashka11, line 362: We also have native inputs available

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  const radiusVal = document.createElement('span');
+  radiusVal.style.cssText = [
+    'font-size:13px', 'color:#ae5d04', 'font-weight:600', 'width:44px',
+    'text-align:right', 'flex-shrink:0', 'font-family:Consolas,monospace',
+  ].join(';');
+  radiusCtrl.appendChild(radiusSlider);
+  radiusCtrl.appendChild(radiusVal);
+  radiusRow.appendChild(radiusLabel);
+  radiusRow.appendChild(radiusCtrl);
+  popBody.appendChild(radiusRow);
```
Aleksashka11: For proper forms, we already have ui.input.form. For layout, we can use ui.divV and ui.divH.

I strongly recommend checking our API examples here:
[https://github.com/datagrok-ai/public/tree/master/packages/ApiSamples/scripts/ui](https://github.com/datagrok-ai/public/tree/master/packages/ApiSamples/scripts/ui)

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  /** Finds the visible bounding box for the popover — walks up from `parent`
+   *  and intersects every overflow-hidden ancestor's rect. Mol*'s own
+   *  layout wrappers (`.msp-layout-static`, `.msp-layout-region`) can be
+   *  wider than the host Datagrok `<div>`, so taking only the first
+   *  overflow:hidden ancestor would miss the tighter clip further up.
+   *  @param {HTMLElement} parent element to start the walk from.
+   *  @return {ClipBox} viewport-relative clip box (inclusive bounds). */
+  type ClipBox = {left: number; right: number; top: number; bottom: number; width: number};
+  const findClipBox = (parent: HTMLElement): ClipBox => {
+    let minLeft = -Infinity; let minTop = -Infinity; let maxRight = Infinity; let maxBottom = Infinity;
```
Aleksashka11: Ouch — that should definitely be split across multiple lines

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+    const width = Math.max(180, Math.min(naturalWidth, availableWidth));
+
+    popover.style.width = `${width}px`;
+    popover.style.right = `${rightOffset}px`;
+    popover.style.left = 'auto';
+
+    // Top: align with the Mol* strip's top when present, else 10 px.
+    const strip = parent.querySelector('.msp-viewport-controls-buttons') as HTMLElement | null;
+    const stripTop = strip ? offsetTopWithin(strip, parent as HTMLElement) : null;
+    const top = stripTop !== null ? stripTop : 10;
```
Aleksashka11: This can be made a bit cleaner and more idiomatic using nullish coalescing:
const top = stripTop ?? 10;

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  // Prevent focus-steal from Mol* canvas on mousedown.
+  iconBtn.addEventListener('mousedown', (e: MouseEvent) => { e.preventDefault(); });
+  iconBtn.addEventListener('click', (e: MouseEvent) => {
+    e.stopPropagation();
+    if (popoverOpen) closePopover(); else openPopover();
+  });
+  iconBtn.addEventListener('keydown', (e: KeyboardEvent) => {
+    if (e.key === 'Enter' || e.key === ' ') {
+      e.preventDefault();
+      if (popoverOpen) closePopover(); else openPopover();
```
Aleksashka11: This shouldn’t be a one-liner
There are quite a few similar cases across this pr — I’d suggest doing a pass and review the code one more time

packages/BiostructureViewer/src/tests/binding-site-tests.ts
talagayev: Ah damn yes, forgot that Claude is a fan of using any a lot, was the case for the last PR that I needed to remind him to not use it. Will adjust it.

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
drizhina: while I agree generally to this point, anything up to 10-50K is virtually unnoticable :)

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
drizhina: here its fine imho. much more muscle flex for one line

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
drizhina: here I agree. there is a class and should be moved to css.

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+    'flex-direction:column',
+    'align-items:flex-end',
+    'pointer-events:none',
+  ].join(';');
+
+  // Icon container holds a semi-transparent backdrop behind the button so the
+  // icon blends with the canvas the same way the native strip icons do.
+  const iconHolder = document.createElement('div');
+  iconHolder.className = 'bsv-bs-icon-holder';
+  iconHolder.style.cssText = [
```
drizhina: and here

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
drizhina: yes. lets also make sure CC knows about our ui library
ui.icon, ui.iconFA, ui.button, etc.

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  // inherit:
+  //   msp-btn-link-toggle-off → color: #9c835f (muted, default)
+  //   :hover                  → color: #ae5d04 (orange accent)
+  //   msp-btn-link-toggle-on  → color: #332b1f (fully saturated on active)
+  iconBtn.className = 'msp-btn msp-btn-icon msp-btn-link msp-btn-link-toggle-off bsv-bs-icon-btn';
+  iconBtn.type = 'button';
+  iconBtn.setAttribute('aria-label', 'Binding site');
+  iconBtn.setAttribute('title', 'Binding site');
+  iconBtn.innerHTML = BULLSEYE_SVG;
+  iconBtn.style.cssText = [
```
drizhina: and here

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  wrapper.appendChild(iconHolder);
+
+  // Popover attached to wrapper.parentElement (viewer root). Styled to
+  // match Mol*'s own `.msp-viewport-controls-panel` — FLAT: no border,
+  // no border-radius, no shadow. Only bg #e0ddd4 and a subtle darker header.
+  const popover = document.createElement('div');
+  popover.className = 'msp-viewport-controls-panel bsv-bs-popover bsv-bs-popover-hidden';
+  popover.setAttribute('role', 'dialog');
+  popover.setAttribute('aria-label', 'Binding site options');
+  popover.style.cssText = [
```
drizhina: and here :D :D

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
drizhina: Yes. its a matter of knowing what we have available :D :D 
for your info, we have ui.iconFa which uses fa icons. we have that loaded on DG, so anything you can find on font awesonme site, can also be used directly, and sizes, colors, the lot can be modified through styles.
check this:
https://fontawesome.com/v4/icon/bullseye
this would be just ui.iconFA('bullseye', handler, tooltip)

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
drizhina: ui.input.slider for example, with bunch of properties like min, max, step, etc. but I guess to your point @Aleksashka11 , these inputs need to match the styles of MOL* , not datagrok. so it is fine. I think its ok here.

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
```diff
@@ -0,0 +1,735 @@
...
+  showChk.style.cssText = 'width:13px;height:13px;accent-color:#ae5d04;cursor:pointer;margin:0';
+  showCtrl.appendChild(showChk);
+  showRow.appendChild(showLbl);
+  showRow.appendChild(showCtrl);
+  popBody.appendChild(showRow);
+
+  // Row: radius slider
+  const radiusRow = document.createElement('div');
+  radiusRow.style.cssText = ROW_STYLE;
+  const radiusLabel = document.createElement('span');
```
drizhina: for most of these guys, its easier to use ui.span, ui.div and most of properties that take bunch of lines take now one line. same for inputs bellow and etc. min/max/step are properties of this input. tell CC to investigate how other plugins do it and have a look inside js-api/ui.ts

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
drizhina: don't really agree here. why. this is a standard and quite good way to create bunch of vars. Actually stemming from CPP and its conventions :D but up to you

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
drizhina: just FYI about how JS does that
In a lovely land of JS
null === undefined => false
null == undefined => true
''(empty string), null, undefined, 0, empty are all nullish
also, while ?? is indeed called nullish coalescing, `'' ?? 10 will give ''`

one that reacts to all nullish values is ||

'' || 10 will give 10.
both null and undefined values will react to ?? 
null ?? undefined ?? 10 => 10

drizhina (commented):
Overall looks good. most comments about usage of better components and code style. 
2 things I will ask you to check
1: make sure it works in normal molstar viewer (as opposed to in context menu). where structures and ligands are there as well.
2: check for any memory leaks. I fixed tons of memory leaks about 2 years ago, and know that it is super easy to accumulate stuff without knowing. I can help you with it tomorrow

talagayev:
> Overall looks good. most comments about usage of better components and code style. 2 things I will ask you to check 1: make sure it works in normal molstar viewer (as opposed to in context menu). where structures and ligands are there as well. 2: check for any memory leaks. I fixed tons of memory leaks about 2 years ago, and know that it is super easy to accumulate stuff without knowing. I can help you with it tomorrow

Yes the 1st one I can do, the second one would be nice to see how you approached it :)

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
Aleksashka11: There are many more examples like this in the code — this was just the first one I came across. A single line is fine, but further down it’s far more than that...

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
Aleksashka11: They’ll match anyway since we’re layering a lot of custom styling on top. I’d rather not have the codebase filled with document.createElement in places where it’s not really needed

Our code is what we feed to LLMs — if this is the pattern we provide, it’s the same pattern we’ll get back
I would really avoid that if that is possible and here I feel that it can be the case

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
Aleksashka11: Ah, those CPP guys 😄

skalkin: a slightly shorter one? :)

let minLeft = -Infinity, minTop = -Infinity, maxRight = Infinity, maxBottom = Infinity;

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
skalkin: Yes, setup linter or rules for CC. Actually we might want to have them in the central place.

packages/BiostructureViewer/src/viewers/molstar-viewer/binding-site.ts
talagayev: ah yes that would be a good idea :)

PR #3783, opened 2026-05-06
BiostructureViewer: Protein-ligand interaction implementation

talagayev:
Implements [ProLIF](https://github.com/chemosim-lab/ProLIF) for protein-ligand interactions, which allows to analyze the interactions between the protein-ligand complex.


packages/BiostructureViewer/scripts/protein_ligand_interactions.py
```diff
@@ -0,0 +1,489 @@
...
+def is_pdbqt(text):
+    # Ligand PDBQT marker records
+    if 'ROOT' in text or 'BRANCH' in text or 'TORSDOF' in text:
+        return True
+    # Receptor PDBQT: no ROOT/BRANCH but ATOM/HETATM lines carry AutoDock atom
+    # types in cols 78-79 (OA, NA, HD, NS, SA, HS, NX, OS) instead of standard
+    # element symbols. Sample the first 50 ATOM/HETATM lines.
+    ad_specific = {'OA', 'NA', 'HD', 'NS', 'NX', 'OS', 'SA', 'HS'}
+    checked = 0
+    for line in text.splitlines():
```
Aleksashka11: I don’t love that we’re calling splitlines() on the same text multiple times across different methods. Ideally, we’d do it once, cache the result, and reuse it everywhere (as an example)

splitlines() is already O(n) operation, and right now we’re effectively turning that into O(n × k) depending on how many times we call it

packages/BiostructureViewer/scripts/protein_ligand_interactions.py
```diff
@@ -0,0 +1,489 @@
...
+            if os.path.exists(path):
+                os.unlink(path)
+            return
+        except OSError:
+            time.sleep(0.1)
+
+
+def is_pdbqt(text):
+    # Ligand PDBQT marker records
+    if 'ROOT' in text or 'BRANCH' in text or 'TORSDOF' in text:
```
Aleksashka11: There’s a more pythonic way to do it, e.g. using any()
But as you prefer — I think this is mostly a matter of taste

packages/BiostructureViewer/scripts/protein_ligand_interactions.py
```diff
@@ -0,0 +1,489 @@
...
+        return True
+    # Receptor PDBQT: no ROOT/BRANCH but ATOM/HETATM lines carry AutoDock atom
+    # types in cols 78-79 (OA, NA, HD, NS, SA, HS, NX, OS) instead of standard
+    # element symbols. Sample the first 50 ATOM/HETATM lines.
+    ad_specific = {'OA', 'NA', 'HD', 'NS', 'NX', 'OS', 'SA', 'HS'}
```
Aleksashka11: Btw, we can also move the constants outside the function so they aren’t recreated on every call

packages/BiostructureViewer/scripts/protein_ligand_interactions.py
```diff
@@ -0,0 +1,489 @@
...
+    checked = 0
+    for line in text.splitlines():
+        if line[:6] in ('ATOM  ', 'HETATM') and len(line) >= 79:
```
Aleksashka11: Just a question — why can’t we use startswith here?

packages/BiostructureViewer/scripts/protein_ligand_interactions.py
```diff
@@ -0,0 +1,489 @@
...
+                break
+    return False
+
+
+def pdbqt_to_pdb(text):
+    """Strip AutoDock-specific records and restore element symbols in cols 77-78."""
+    out = []
+    for line in text.splitlines():
+        h = line[:6]
+        if h.startswith(('ROOT', 'ENDROOT', 'BRANCH', 'ENDBRA', 'TORSDO')):
```
Aleksashka11: Oh, ok, here we use startswith

But I’m not sure I understand why we first do slicing. The result should be the same whether we slice first and do startswith or just use startswith directly

packages/BiostructureViewer/scripts/protein_ligand_interactions.py
```diff
@@ -0,0 +1,489 @@
...
+                ch, rn = L[21], int(L[22:26])
+                if ch not in min_per_chain or rn < min_per_chain[ch]:
+                    min_per_chain[ch] = rn
+            except ValueError:
+                pass
+    offsets = {ch: (1 - mn) for ch, mn in min_per_chain.items() if mn < 1}
+    if not offsets:
+        return pdb_text
+    out = []
+    for L in pdb_text.splitlines():
```
Aleksashka11: Iterate twice over the same data

packages/BiostructureViewer/scripts/protein_ligand_interactions.py
```diff
@@ -0,0 +1,489 @@
...
+    """Concat protein + ligand into one PDB, with TER between."""
+    p = [L for L in protein_pdb.splitlines() if not L.startswith('END')]
+    l = [L for L in ligand_pdb.splitlines()
+         if L[:6] in ('ATOM  ', 'HETATM', 'CONECT')]
+    return '\n'.join(p + ['TER'] + l + ['END'])
+
+
+def detect_ligand_resname(pdb_text):
+    counts = Counter(L[17:20].strip() for L in pdb_text.splitlines()
+                     if L.startswith('HETATM') and L[17:20].strip() not in SKIP_RESNAMES)
```
Aleksashka11: L[17:20].strip() is computed twice
I also don’t like that everything implicitly depends on fixed-width slicing without safeguards — at minimum, we should add a length check I think

packages/BiostructureViewer/scripts/protein_ligand_interactions.py
```diff
@@ -0,0 +1,489 @@
...
+
+# 4. pdbfixer for protein hydrogens. Speed optimization: shrink the protein
+# to just the binding-site neighborhood (8 Å around the ligand) before running
+# pdbfixer. Hydrogenation scales linearly with atom count, so for a 300-residue
+# protein this is typically a 5-10× speedup. We only do this when a ligand is
+# available to define the neighborhood — for the rare ligand-less case (which
+# shouldn't reach this script anyway) the full protein is kept.
+fd_in, in_path = tempfile.mkstemp(suffix='.pdb'); os.close(fd_in)
+fd_out, fixed_path = tempfile.mkstemp(suffix='.pdb'); os.close(fd_out)
+fd_combined, combined_path = tempfile.mkstemp(suffix='.pdb'); os.close(fd_combined)
```
Aleksashka11: Let’s at least try to eliminate the need for reading and writing temporary files — file I/O is slow

Aleksashka11: Hm, we could switch to NamedTemporaryFile (if we stick with this approach). That would remove the need for os.close

packages/BiostructureViewer/scripts/protein_ligand_interactions.py
```diff
@@ -0,0 +1,489 @@
...
+    # etc.). Try progressively more permissive modes:
+    #   1. force=True   — skip strict valence check after bond inference
+    #   2. NoImplicit=False — skip bond inference entirely; use implicit Hs.
+    #      The molecule's SMARTS matching still works because heavy-atom
+    #      connectivity is preserved; only bond orders are simpler.
+    try:
+        protein_mol_obj = plf.Molecule.from_mda(protein_ag, force=True)
+    except Exception as e_force:
+        print(f'[ProLIF] from_mda(force=True) failed: {type(e_force).__name__}: {e_force}')
+        print('[ProLIF] Retrying with NoImplicit=False (skip bond-order inference)')
```
Aleksashka11: We can switch to using logging instead of print — that would be cleaner

packages/Docking/src/package.ts
```diff
@@ -18,6 +18,171 @@ import { addColorCoding, buildComparisonTable, formatColumns, getRemarksFromPdb,
...
+      const chain = (line[21] || '').trim() || 'A';
+      const resid = line.slice(22, 26).trim();
+      if (rn && !PROLIF_SKIP_RESNAMES.has(rn)) {
+        const key = `${rn} ${chain} ${resid}`;
+        seen.set(key, (seen.get(key) || 0) + 1);
+      }
+    }
+  }
+  return Array.from(seen.entries()).sort((a, b) => b[1] - a[1]).map(([k]) => k);
+}
```
Aleksashka11: We should definitely move this duplicated logic into a shared place
@drizhina might have a better idea of where it should live in the codebase

packages/BiostructureViewer/css/viewer.css
```diff
@@ -32,4 +32,29 @@
...
     background-size: contain;
-}
\ No newline at end of file
+}
+
+/* "Compute for whole dataset" button in the PL context-panel widget */
+.pl-batch-btn {
+    margin-top: 12px;
+    justify-content: flex-start;
+}
```
Aleksashka11, line 41: Typically, we use the package name as a prefix — for example, bsv-pl-batch-btn

packages/BiostructureViewer/scripts/protein_ligand_interactions.py
```diff
@@ -0,0 +1,892 @@
...
+n_hydrophobic = type_counts.get('HY', 0)
+n_pistacking = type_counts.get('PI', 0)
+n_cationic = type_counts.get('CAT', 0)
+n_anionic = type_counts.get('AN', 0)
+n_cationpi = type_counts.get('CPI', 0)
+n_xbond = type_counts.get('XB', 0)
+n_metal = type_counts.get('M', 0)
+n_vdw = type_counts.get('VDW', 0)
+n_total = sum(type_counts.values())
+
```
Aleksashka11, line 1029: Just to avoid calling .get() repeatedly, we can define a mapping and iterate once
The df construction can be simplified in the same way but up to you :sweat_smile:

packages/BiostructureViewer/src/utils/prolif-panel.ts
```diff
@@ -0,0 +1,406 @@
...
+  const body = ui.div();
+  host.append(body);
+
+  const detectionSource = (params.ligand && params.ligand.trim()) || params.protein;
+  const ligands = detectNonWaterHetatmInstances(detectionSource);
+
+  const compute = (resname: string, target: HTMLElement) => {
+    ui.empty(target);
+    const loader = ui.loader();
+    target.append(loader);
```
Aleksashka11: You can also check other examples of loading indicators here:
https://github.com/datagrok-ai/public/blob/6f0234e388929a5ec6b4f2924febe96faff64623/packages/ApiSamples/scripts/ui/components/loading-indicators.js#L11

Just for reference what else you can use

packages/BiostructureViewer/src/utils/prolif-panel.ts
```diff
@@ -0,0 +1,406 @@
...
+
+  if (ligands.length === 0 && !params.ligand_resname)
+    body.append(ui.divText('No non-water HETATM ligand found in this structure.'));
+  else if (params.ligand_resname || ligands.length === 1)
+    compute(params.ligand_resname || ligands[0], body);
+  else {
+    const innerBody = ui.div();
+    const picker = ui.input.choice('Ligand', {
+      value: null, items: ligands, nullable: true,
+      onValueChanged: (v: string | null) => { if (v) compute(v, innerBody); },
```
Aleksashka11: Just a minor comment — I don’t think this is very readable as-is
I would avoid writing it as a single line :sweat_smile:

packages/BiostructureViewer/src/utils/prolif-panel.ts
```diff
@@ -0,0 +1,406 @@
...
+
+  // Always-rendered batch button. Disabled with tooltip when batchCtx missing.
+  const btn = ui.button(
+    'Compute for whole dataset',
+    () => batchCtx != null ? runPlBatchForDataset(batchCtx) : undefined,
+  );
+  btn.classList.add('pl-batch-btn');
+  if (batchCtx == null) {
+    btn.setAttribute('disabled', '');
+    btn.title = 'No source DataFrame context — open this panel from a grid cell.';
```
Aleksashka11: I’m not entirely sure, but I think we should use Datagrok tooltips here

packages/BiostructureViewer/src/package.ts
```diff
@@ -229,6 +239,68 @@ export class PackageFunctions {
...
+    let pdbCol: DG.Column<string> | null = null;
+    if (hasValidCell) {
+      df = cell.dataFrame;
+      pdbCol = cell.column as DG.Column<string>;
+    } else {
+      // Fallback: use the current table view's DataFrame. Lets the button
+      // work even when the panel is invoked without a usable cell context.
+      const t = grok.shell.t;
+      if (t != null) {
+        df = t;
```
Aleksashka11: I might be missing something, but why is that?

packages/BiostructureViewer/src/package.ts
```diff
@@ -229,6 +239,68 @@ export class PackageFunctions {
...
+        const m3dCols = t.columns.toList()
+          .filter((c) => c.semType === 'Molecule3D' || c.tags['quality'] === 'Molecule3D');
```
Aleksashka11: Can’t we just use t.columns.bySemTypeAll(DG.SEMTYPE.MOLECULE3D)?

packages/BiostructureViewer/src/utils/prolif-panel.ts
```diff
@@ -0,0 +1,406 @@
...
+  // For each kept column, write counts for OK rows and DG.INT_NULL for ERROR
+  // rows so per-column stats / sorts / filters distinguish "failed" from
+  // "0 interactions". SKIP and PENDING rows stay at the addNewInt default
+  // (DG.INT_NULL, displayed as empty).
+  for (const c of PL_INT_COLS) {
+    const arr = counts[c.key];
+    let hasAny = false;
+    for (let i = 0; i < rowCount; i++) {
+      if (rowState[i] === ROW_OK && arr[i] > 0) { hasAny = true; break; }
+    }
```
Aleksashka11: Let’s probably update our eslint rules to disallow things like this

packages/BiostructureViewer/src/utils/prolif-panel.ts
Aleksashka11: Also, I prefer btn.disabled = true/false much more

libraries/bio/src/prolif/prolif-panel.ts
```diff
@@ -0,0 +1,701 @@
+// Protein-Ligand interaction (ProLIF) context panel + batch handler.
```
drizhina, line 1: prolif panels and all need to be unified in BSV package, no need to have it in docking separately. therefore no need to have this file in bio library, should be there as well.

libraries/bio/src/prolif/pl-png-cell-renderer.ts
```diff
@@ -0,0 +1,98 @@
...
+  /** Datagrok's true LRU (proper access-order eviction, not just insertion-
+   *  order). Maps a stable `(dfId, colName, rowIndex, colVersion)` key to
+   *  the decoded HTMLImageElement so subsequent renders are synchronous.
+   *  Capacity is 200 — at the configured row height (130px) the visible
+   *  grid is rarely more than ~40 rows × a few diagram columns, so 200
+   *  comfortably covers the visible viewport plus scroll-back. */
+  private readonly _cache: DG.LruCache<string, HTMLImageElement> =
+    new DG.LruCache<string, HTMLImageElement>(200);
+
+  render(g: CanvasRenderingContext2D, x: number, y: number, w: number, h: number,
```
drizhina: 1. Bio lib is definitely not the place for this. any renderer needs to be in package.
2. sorry but you reinvented this :D :D please look into powergrid/src/png-renderer.ts :D :D 
if you set semtype to rawPng it will work out of the box

drizhina (changes requested):

PR #3784, opened 2026-05-07
[report 22615] CLAUDE-132: bio: Guard Bio:SequenceSpaceEditor against tables without Macromolecule columns

PavloPolovyi:
## Summary

Fix `NullError: method not found: 'name' on null` thrown when invoking `Bio | Analyze | Sequence Space...` on a table without any Macromolecule columns.

## Root cause

`DimReductionBaseEditor`'s constructor (in `@datagrok-libraries/ml/src/functionEditors/dimensionality-reduction-editor.ts:166`) reads `this.colInput.value!.name` without a null guard. When the current table has no column matching the requested semtype (`Macromolecule`), `getColInput` returns a column input whose value is `null`, and the constructor throws.

## Fix

Guard the entry point `Bio.SequenceSpaceEditor` (in `public/packages/Bio/src/package.ts`): if the current table view has no Macromolecule column, surface a friendly `grok.shell.error` and return before constructing the editor, instead of crashing inside the bundled package.

## Crash site

- Package: Bio (crash surfaces from bundled `dist/package.js` of Bio)
- Underlying file: `@datagrok-libraries/ml/src/functionEditors/dimensionality-reduction-editor.ts`, `DimReductionBaseEditor` constructor (~line 166)
- Error pattern: `NullError: method not found: 'name' on null` (surfaces in JS console as `TypeError: Cannot read properties of null (reading 'name')`)

## Testing

- `npm run build` in `public/packages/Bio` — exits 0 (only pre-existing asset-size warnings).
- Reproduction recipe (`grok.functions.call('Bio:SequenceSpaceEditor', {})` on a `demog` view) used to throw the NullError; with this guard, the user now sees a clear error toast instead of an unhandled exception.

### DIFF
--- a/packages/Bio/CHANGELOG.md
+++ b/packages/Bio/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Bio changelog
 
+## v.next
+
+### Bug Fixes
+
+* Sequence Space: Show a friendly error instead of crashing when invoked on a table without Macromolecule columns
+
 ## 2.27.4 (2026-04-29)
 
 * Support for non canonical RNA and correct splitting in helm
--- a/packages/Bio/src/package.ts
+++ b/packages/Bio/src/package.ts
@@ -236,6 +236,11 @@ export class PackageFunctions {
 
   @grok.decorators.editor({tags: ['editor']})
   static SequenceSpaceEditor(call: DG.FuncCall) {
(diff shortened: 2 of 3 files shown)


LaykDimon:
unneeded package-lock is here + the fix is not thast good + fixed on master

PR #3789, opened 2026-05-08
CLAUDE-159: KetcherSketcher: Defer Ketcher React Editor mount until ketcher-host has non-zero dimensions, using a ResizeObserver to avoid SVGLength relative-length resolution failure in <RulerArea>

PavloPolovyi:
## Summary

Fix `Failed to read the 'value' property from 'SVGLength': Could not resolve relative length` thrown from ketcher-react's `<RulerArea>` when the KetcherSketcher widget is mounted into a hidden or zero-sized container — observed when expanding the Chemistry → Gasteiger Partial Charges info panel on a smiles cell, and reproducible by mounting `grok.functions.call('KetcherSketcher:ketcherSketcher', {})` into a `width:0;height:0;overflow:hidden` div.

## Root cause

ketcher-react's macromolecules `<EditorContainer>` renders its canvas with `width="100%" height="100%"` and `<RulerArea>` reads `editor.canvas.width.baseVal.value` on mount (`vendors-node_modules_ketcher-react_dist_index_modern-b9db71c5_js.js:20130`, source ≈ `index.modern-b9db71c5.js:18843`). Reading `.baseVal.value` on a percentage-based `SVGLength` throws `NotSupportedError` when the SVG's containing block has no resolvable width.

`KetcherSketcher`'s constructor in `public/packages/KetcherSketcher/src/ketcher.tsx` was calling `ReactDOM.createRoot(this.ketcherHost).render(<Editor .../>)` synchronously, before `this.root` was attached to a sized parent. When React committed, the Editor (and its `<RulerArea>`) mounted against a host whose layout was still zero-sized, and the SVG could not resolve relative units.

## Fix

Defer the React Editor mount until `ketcherHost` actually has non-zero dimensions:

- Construct the React element eagerly, but do not call `ReactDOM.createRoot(...).render(...)` yet.
- Check `getBoundingClientRect()` synchronously; if the host already has a real size (the common in-grid / sketcher-dialog case), mount immediately — behavior unchanged.
- Otherwise, observe the host with a `ResizeObserver`; mount the Editor on the first observation with width > 0 and height > 0, then disconnect.
- `detach()` disconnects the observer and clears the cached element so a hidden-but-never-shown widget is cleaned up correctly.

No changes to ketcher-react, ketcher-core, or sibling packages — the fix is contained to KetcherSketcher.

## Testing

- `npm run build` (KetcherSketcher) exits 0.
- Reproduced original error against the unfixed package: console fired `NotSupportedError: Failed to read the 'value' property from 'SVGLength': Could not resolve relative length.` from `<RulerArea>` immediately on mount into a zero-sized div, and on the Gasteiger Partial Charges info-panel path.
- With the fix applied, `<RulerArea>` is no longer mounted while the host is zero-sized, so the failing `editor.canvas.width.baseVal.value` line cannot execute against an unresolvable SVG. When the host is later sized, the Editor mounts normally.

Reproduction evidence is attached to the JIRA comment thread (smiles-grid + gasteiger-ruler-error screenshots).

### DIFF
--- a/packages/KetcherSketcher/CHANGELOG.md
+++ b/packages/KetcherSketcher/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Ketcher Sketcher changelog
 
+## 2.4.3 (2026-05-08)
+
+* KetcherSketcher: Defer React Editor mount until the host has non-zero dimensions, preventing RulerArea SVGLength `Could not resolve relative length` error when the sketcher is rendered into a hidden or zero-sized container
+
 ## 2.4.2 (2026-04-23)
 
 * Reset explicitMol on change only if molecule has been already set
--- a/packages/KetcherSketcher/package-lock.json
+++ b/packages/KetcherSketcher/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@datagrok/ketcher-sketcher",
-  "version": "2.4.2",
+  "version": "2.4.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@datagrok/ketcher-sketcher",
-      "version": "2.4.2",
+      "version": "2.4.3",
       "dependencies": {
         "@datagrok-libraries/test": "^1.1.0",
         "@datagrok-libraries/utils": "^4.6.5",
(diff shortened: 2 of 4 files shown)


PavloPolovyi:
@MariaDolotova be skeptical when reviewing this)

MariaDolotova:
The changes are valid and were included in this PR - https://github.com/datagrok-ai/public/commit/751b8aff771f329ee95653dd8d1aee3a7d896d67

PR #3801, opened 2026-05-13
[ticket] GROK-20104: Configure auth: Drop Google/Facebook/GitHub sections

PavloPolovyi:
The dedicated Google/Facebook/GitHub OAuth login methods were removed from Datagrok core in favour of OpenID Connect. Update the user-facing setup doc accordingly: drop the OAuth section, broaden the OpenID section to mention Google/Azure AD/Okta/Auth0 explicitly, and point Google users at the OIDC discovery URL.


LaykDimon:
this is a better one and actually rewrites what was removed - merging

PR #3804, opened 2026-05-14
Chem: Pharmacophore feature highlighting

talagayev:
Changes done:

- Added column with molecule containing the colored pharmacophore features


packages/Chem/src/panels/pharmacophore-features.ts
```diff
@@ -47,18 +50,42 @@ export async function runPharmacophoreDetection(
...
+    const letter = familyCol.get(i);
+    const smartsVal = smartsCol.get(i);
+    if (!letter || !smartsVal) continue;
+    const familyId = FAMILY_LETTER_TO_ID[letter];
+    const info = FAMILY_INFO[letter];
+    if (familyId && info && familySet[familyId])
+      scaffolds.push({color: info.color, molecule: smartsVal});
+  }
+  const clone = moleculeCol.clone();
+  clone.name = 'Pharmacophore';
```
drizhina: you should use moleculeCol.dataFrame.columns.getUnusedName(`Pharmacophore(${moleculeCol.name})`) to indicate which column this is targeting, and there are no clashes in names

drizhina: Oh, and also, if you notice, most stuff in that table is yellow, and features that are less specific and match on many parts. can you sort the features table (when highlighting) so that more colors show up, for example, sort by number of appearances of that color, so that more would end up first, and less last.

PR #3805, opened 2026-05-14
[report 22120] CLAUDE-533: Chem: Guard against null mmpViewer after awaitCheck timeout in MMP demo before setting helpUrl

PavloPolovyi:
## Summary

Fix `TypeError: Cannot set properties of null (setting 'helpUrl')` at `packages/Chem/src/demo/demo.ts` (`_demoMMPA`) when the Matched Molecular Pairs demo view is closed before its MMP viewer finishes registering.

## Root cause

In `_demoMMPA`, `mmpViewer` is initialised to `null` and only assigned inside an `awaitCheck` callback that polls `tv.viewers` for an `'Matched Molecular Pairs Analysis'` entry. The `awaitCheck` has a 20s timeout, and its surrounding `try/catch` swallows the timeout error with `catch (e) {};`. The following line then performs `mmpViewer!.helpUrl = '...'`, using a non-null assertion that masked the unsafe assumption: if the user closed the demo view (or it never registered an MMP viewer within 20s), `mmpViewer` remained `null` and the assignment threw.

Reproduced via the path in the user report: open Browse → Apps → Demo → Cheminformatics → Matched Molecular Pairs, close the demo view before the MMP viewer is added to `tv.viewers`, then open another file. After ~20s the error `TypeError: Cannot set properties of null (setting 'helpUrl')` fires with stack at `webpack://chem/src/demo/demo.ts ... helpUrl`, matching the error pattern from triage.

## Fix

After the `try/catch`, check whether `mmpViewer` is `null` and bail out early — also skipping the dependent `setTimeout` that calls `grok.shell.windows.help.showHelp`, which is meaningless once the demo view is gone. A `MatchedMolecularPairsViewer | null` cast is required because TypeScript narrows the closure-assigned variable to `never` after the null check.

## Testing

- `npm run build` in `packages/Chem/` exits 0 (only pre-existing asset-size warnings).
- Verified the failing line `mmpViewer!.helpUrl = ...` (formerly line 211) is no longer reachable with `mmpViewer === null`; the same reproduction recipe that produced the original `TypeError` now exits the demo helper cleanly when the view is closed mid-init.
- Added a `## v.next` CHANGELOG entry.

### DIFF
--- a/packages/Chem/CHANGELOG.md
+++ b/packages/Chem/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## v.next
 
 * GROK-20108: Docker API: Throw a clear error when the `chem-chem` container is not available, instead of `Cannot read properties of undefined (reading 'id')`
+* MMP demo: Guard against null `mmpViewer` after the 20s `awaitCheck` timeout (e.g. when the user closes the demo view before the viewer registers), preventing `TypeError: Cannot set properties of null (setting 'helpUrl')`
 
 ## 1.17.8 (2026-05-12)
 
--- a/packages/Chem/package-lock.json
+++ b/packages/Chem/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@datagrok/chem",
-  "version": "1.17.6",
+  "version": "1.17.8",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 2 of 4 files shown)


LaykDimon:
fixed + the fix here is not that good as on master

Issue #1301, opened 2022-12-06
Chem: Similarity Search: Sketch: a "search as you sketch" option in the bottom


skalkin:

skalkin:
It would make sense to add the "Search as you sketch" checkbox in the dialog, below the sketcher.

dnillovna:
This issue has been mirrored in Jira: https://reddata.atlassian.net/browse/GROK-14840

dnillovna:
This issue has been mirrored in Jira: https://reddata.atlassian.net/browse/GROK-16095

LesiaPavlenko:
Chem > 1.17.9

PR #3812, opened 2026-05-15
Chem-1301 : Similarity Search: Sketch: a "search as you sketch" option

talagayev:
Changes made:

- Addition of dynamic molecule update while sketching in `Similarity Search`


drizhina (approved):
LGTM. just fyi, when targetting tickets from github, the PR name should be #ticketnum:  description.
like 
#1301: Similarity search.....
but no problem at thes time. this just does the automatic asignment of commit to ticket. feel free to merge

MariaDolotova:
Just a small refactoring suggestion (not critical, so up to you) There are duplicated 3 rows in 3 places. Can extract into one small function. 
image

MariaDolotova (approved):

talagayev:
> Just a small refactoring suggestion (not critical, so up to you) There are duplicated 3 rows in 3 places. Can extract into one small function. image

Ah yes true, agree would be more cleaner then to have the duplicate in two places, will adjust it :)

PR #3814, opened 2026-05-18
GROK-12835: Chem: R-group Analysis: Add the ability to break the execution of the function

talagayev:
The function itself must have a cancellation mechanism, for example, a break of a loop by condition.

Changes made:
- Added execution canceling for R-Groups Analysis


drizhina (approved):
lgtm

Aleksashka11 (approved):

PR #3826, opened 2026-05-27
GROK-15374: Structure Filter: update of cell doesn't lead to row being filtered out

talagayev:
Changes made:

- Adjusted code to filter out the molecules when modified through sketching

### DIFF
--- a/packages/Chem/src/widgets/chem-substructure-filter.ts
+++ b/packages/Chem/src/widgets/chem-substructure-filter.ts
@@ -468,9 +468,18 @@ export class SubstructureFilter extends DG.Filter {
         this.recalculateFilter = true;
         return;
       }
-      // no cached result yet, or a full search is running -> recompute the whole column
-      const searching = !!this.batchResultObservable && !this.batchResultObservable.closed;
-      if (!this.bitset || this.recalculateFilter || searching) {
+      // A full search is already running: don't restart it (avoids a restart storm when several
+      // cells are edited in quick succession). Just mark dirty — when the search completes,
+      // applyFilter sees the flag and recomputes once with the edited rows included.
+      if (this.batchResultObservable && !this.batchResultObservable.closed) {
+        this.recalculateFilter = true;
+        return;
+      }
+      // Recompute the whole column instead of patching when there's no cached result, a recompute
+      // is already pending, or EXACT_MATCH (the worker compares raw strings against the canonicalized
+      // column, which a raw edited cell value can't reproduce — see searchWithPatternFps).
+      if (!this.bitset || this.recalculateFilter ||
+        this.searchType === SubstructureSearchType.EXACT_MATCH) {
         this.recalculateFilter = true;
         await this._onSketchChanged();
         return;
--- a/packages/Chem/src/widgets/chem-substructure-filter.ts
+++ b/packages/Chem/src/widgets/chem-substructure-filter.ts
@@ -62,6 +62,11 @@ type AlignHighlightSync = {
   tableName: string;
 }
 
+/** The parts of the (untyped) onValuesChanged payload this filter reads. */
+interface ValuesChangedEventArgs {
+  args?: {column?: DG.Column; indexes?: Int32Array | number[]};
+}
+
 export class SubstructureFilter extends DG.Filter {
   // @ts-ignore
   sketcher: DG.chem.Sketcher = new DG.chem.Sketcher();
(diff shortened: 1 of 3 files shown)


packages/Chem/src/widgets/chem-substructure-filter.ts
```diff
@@ -289,6 +289,22 @@ export class SubstructureFilter extends DG.Filter {
         }
       }));
 
+    // Re-run the search when cells in our column are edited (e.g. via the in-cell sketcher);
+    // otherwise the cached bitset stays stale and edited rows keep their old match state.
+    // Only the active filter recomputes — peers sync via FILTER_SYNC_EVENT.
+    this.subs.push(this.dataFrame!.onValuesChanged
```
MariaDolotova, line 315: I'm not sure I like the concept of recalculating filter for the whole column based on onValuesChanged event. Will try to explain my point of view :) 

Let's assume we have a column of 1 million molecules. And most likely, user will change only one molecule at once (using cell editor).

And changing this one molecule will force filter recalculation for the whole million (via calling this._onSketchChanged()). A better way is just to check if the changed molecule still satisfies the filter condition and either exclude it from the bitset ot not. I assume you have the indexes of the changed rows - args?.args?.indexes.

packages/Chem/src/widgets/chem-substructure-filter.ts
```diff
@@ -436,6 +461,48 @@ export class SubstructureFilter extends DG.Filter {
     grok.shell.tv?.grid?.invalidate();
   }
 
+  /** Re-evaluate just the changed rows and patch the cached bitset in place.
+   * Safe only for substructure search types — see the onValuesChanged guard. */
+  _patchChangedRows(indexes: Int32Array | number[]): void {
+    const rdkit = getRdKitModule();
```
MariaDolotova: Please consider do this in web-worker

packages/Chem/src/widgets/chem-substructure-filter.ts
```diff
@@ -289,6 +290,30 @@ export class SubstructureFilter extends DG.Filter {
...
+        const active = this.column?.temp[CHEM_APPLY_FILTER_SYNC];
+        return this.isFiltering &&
+          args?.args?.column?.name === this.columnName &&
+          args?.args?.indexes?.length &&
+          (!active || active.filterId === -1 || active.filterId === this.filterId);
+      }))
+      .subscribe(async (args: any) => {
+        const searching = this.batchResultObservable && !this.batchResultObservable.closed;
+        const canPatch = this.bitset != null && !searching &&
+          this.searchType !== SubstructureSearchType.IS_SIMILAR &&
```
Aleksashka11: We don't need to check searchType here

packages/Chem/src/widgets/chem-substructure-filter.ts
```diff
@@ -289,6 +290,30 @@ export class SubstructureFilter extends DG.Filter {
         }
       }));
 
+    // Patch only the changed rows' bits when cells are edited — O(k) instead of full re-scan.
+    // Falls back to a full recompute when the cache can't help (no bitset, search in flight,
+    // or a search mode that needs precomputation: IS_SIMILAR, STEREO_AGNOSTIC).
+    this.subs.push(this.dataFrame!.onValuesChanged
+      .pipe(filter((args: any) => {
+        const active = this.column?.temp[CHEM_APPLY_FILTER_SYNC];
```
MariaDolotova: please also check cases with disablign/enabling the filter

packages/Chem/src/widgets/chem-substructure-filter.ts
```diff
@@ -289,6 +290,30 @@ export class SubstructureFilter extends DG.Filter {
...
+        return this.isFiltering &&
+          args?.args?.column?.name === this.columnName &&
+          args?.args?.indexes?.length &&
+          (!active || active.filterId === -1 || active.filterId === this.filterId);
+      }))
+      .subscribe(async (args: any) => {
+        const searching = this.batchResultObservable && !this.batchResultObservable.closed;
```
MariaDolotova: In case we have running search at the moment we changed some molecule - first wait for the search to be completed, then run search for the changed molecules

packages/Chem/src/widgets/chem-substructure-filter.ts
```diff
@@ -289,6 +290,30 @@ export class SubstructureFilter extends DG.Filter {
...
+        const canPatch = this.bitset != null && !searching &&
+          this.searchType !== SubstructureSearchType.IS_SIMILAR &&
+          this.searchType !== SubstructureSearchType.STEREO_AGNOSTIC;
+        if (canPatch)
+          this._patchChangedRows(args.args.indexes);
+        else {
```
Aleksashka11: No need for the else block
Also, please handle the case when the search is already in progress — await the search completion first, then apply the patch

packages/Chem/src/widgets/chem-substructure-filter.ts
```diff
@@ -289,6 +290,30 @@ export class SubstructureFilter extends DG.Filter {
...
 
+    // Patch only the changed rows' bits when cells are edited — O(k) instead of full re-scan.
+    // Falls back to a full recompute when the cache can't help (no bitset, search in flight,
+    // or a search mode that needs precomputation: IS_SIMILAR, STEREO_AGNOSTIC).
+    this.subs.push(this.dataFrame!.onValuesChanged
+      .pipe(filter((args: any) => {
+        const active = this.column?.temp[CHEM_APPLY_FILTER_SYNC];
+        return this.isFiltering &&
+          args?.args?.column?.name === this.columnName &&
+          args?.args?.indexes?.length &&
```
Aleksashka11: const eventArgs = args?.args;

PR #3827, opened 2026-05-27
Chem: Scope RDKit worker cancellation per operation

talagayev:
Changes made:

- Adjusted code to have for the operation and `opID` to allow the separate operations to be stopped
- Added `opID` for R-Group analysis to be able to stop it, while running in parallel other processes


packages/Chem/src/rdkit-service/rdkit-service-worker-base.ts
```diff
@@ -10,7 +10,19 @@ export class RdKitServiceWorkerBase {
...
+  /** Per-op cancellation set. Lets one op be cancelled without affecting concurrent work. */
+  _terminatedOps: Set<string> = new Set<string>();
   _terminationCheckDelay = 50;
+
+  isOpTerminated(opId?: string): boolean {
+    return opId !== undefined && this._terminatedOps.has(opId);
+  }
+
+  setOpTerminate(opId: string, flag: boolean): void {
+    if (flag) this._terminatedOps.add(opId);
```
Aleksashka11: Should be multiline
Same for the else block

packages/Chem/src/rdkit-service/rdkit-service.ts
```diff
@@ -546,6 +547,17 @@ export class RdKitService {
       });
   }
 
+  async setOpTerminate(opId: string, flag: boolean): Promise<void> {
+    const t = this;
```
Aleksashka11: const t = this is it necessary?

packages/Chem/src/rdkit-service/rdkit-service-worker-base.ts
```diff
@@ -10,7 +10,19 @@ export class RdKitServiceWorkerBase {
   _molsCache: LRUCache<string, RDMol> | null = null;
   _cacheCounter = 0;
   _requestTerminated = false;
+  /** Per-op cancellation set. Lets one op be cancelled without affecting concurrent work. */
+  _terminatedOps: Set<string> = new Set<string>();
   _terminationCheckDelay = 50;
+
+  isOpTerminated(opId?: string): boolean {
+    return opId !== undefined && this._terminatedOps.has(opId);
```
Aleksashka11: Do we expect only undefined here?

packages/Chem/src/tests/menu-tests-rgroups.ts
```diff
@@ -211,6 +212,34 @@ M  END
...
+    let canceled = false;
+    const killed = await runCancellableChemOp(newChemOpId(), () => canceled, async () => {
+      canceled = true; throw new Error('worker terminated');
+    });
+    expect(killed === undefined, true, 'cancelled + throw => undefined');
+    let threw = false;
+    try {
+      await runCancellableChemOp(newChemOpId(), () => false, async () => {throw new Error('real failure');});
+    } catch (_) {threw = true;}
+    expect(threw, true, 'a real error is rethrown, not swallowed');
```
drizhina, line 220: you should test the case where a long task is running like normal, then someone launches another one and cancels the second one. first should not get canceled, and second should.

packages/Chem/src/rdkit-service/rdkit-service.ts
```diff
@@ -556,56 +557,60 @@ export class RdKitService {
...
+      await restart;
+    } finally {
+      if (this._restartPromise === restart)
+        this._restartPromise = null;
+    }
+  }
+
+  // Resolves once any in-progress restartWorker has finished, so callers don't dispatch to a reloading worker.
+  async whenWorkersReady(): Promise<void> {
+    while (this._restartPromise)
```
drizhina: this is incorrect. the restartPromise is for one worker only. why do you even need it?

packages/Chem/src/tests/menu-tests-rgroups.ts
talagayev: Yup can add that test, will do :)

packages/Chem/src/rdkit-service/rdkit-service.ts
talagayev: Here the Idea was to no move the new worker into the pool before `moduleInit` finishes, but yes let me try to adjust this, since I think that could be done with less lines and would be no only for 1 woker

drizhina (approved):

packages/Chem/src/utils/chem-common.ts
```diff
@@ -59,21 +59,69 @@ export function criticalSectionEnd(key: string): void {
...
+  await chemBeginCriticalSection(opId);
+  try {
+    return await work();
+  } finally {
+    chemEndCriticalSection(opId);
+  }
+}
+
+// Cancellable wrapper: undefined if cancelled, else the result.
+export async function runCancellableChemOp<T>(
```
skalkin, line 110: How is it cancelable? I mean - who is canceling the "work()" delegate? I thought the idea was to truly cancel it (stop mid-computation by periodically checking the flag), not just to return null?

PR #3849, opened 2026-06-16
GROK-20222: SequenceTranslator: Markush enumeration - R group addition improvements

talagayev:
Changes made:

- Added option to copy `R*` functional groups from one group to the other
- Added templates for `R*` group generation
- Added option to save `Cores` and `R Group` in a `CSV` file


packages/SequenceTranslator/src/polytool/pt-chem-enum-dialog.ts
```diff
@@ -458,6 +468,193 @@ async function openImportWizard(
...
+  const updateStatus = () => {
+    const t = targetInput.value;
+    if (t == null || !Number.isInteger(t) || t < 1) return fail('Enter a valid R-group number (≥ 1).');
+    if (t === srcN) return fail(`Target R# must differ from the source (R${srcN}).`);
+    const existing = state.rGroupsByNum.get(t)?.length ?? 0;
+    const replace = mergeInput.value === 'Replace';
+    const result = replace ? srcList.length : existing + srcList.length;
+    const noun = srcList.length === 1 ? 'substituent' : 'substituents';
+    const msg = existing > 0 ?
+      `Copy ${srcList.length} ${noun} R${srcN} → R${t} (${replace ? 'replacing' : 'appending to'} ${existing} existing). Result: ${result}.` :
```
drizhina: result variable is not really needed here. when appending, there might be duplicates, so the actual number might be less. just get rid of it

packages/SequenceTranslator/src/polytool/pt-chem-enum-dialog.ts
```diff
@@ -1042,6 +1253,12 @@ function buildChemEnumPanel(
       ui.tooltip.bind(importBtn, `${label}: pick a table + column`);
       parts.push(importBtn);
     }
+    if (onTemplate) {
+      const templateBtn = ui.button('Templates…', onTemplate);
```
drizhina: better use some icon here. like a book or something. we have fant awesome icons, so you can use ui.iconsFa method and provide name and look up the icon on fa icons

packages/SequenceTranslator/src/polytool/pt-chem-enum-dialog.ts
```diff
@@ -1068,9 +1285,14 @@ function buildChemEnumPanel(
   );
   coresClearBtn = coresHeader.clearBtn;
 
+  // Single combined-export button — placed in the Run/footer controls of whichever layout runs.
+  const exportBtn = ui.button('↑ Export', exportAllCsv);
```
drizhina: same, use the download icon here, export is confusing

packages/SequenceTranslator/src/polytool/pt-chem-enum.ts
```diff
@@ -554,10 +690,7 @@ export function enumerateSample(
     for (const [n, rg] of assignment) rgSmiByNum.set(n, rg.smiles);
     const smi = assembleMolecule(core.smiles, rgSmiByNum, rdkit);
     if (!smi) continue;
-
-    const originalRgs = new Map<number, string>();
-    for (const [n, rg] of assignment) originalRgs.set(n, rg.originalSmiles);
-    const item: ChemEnumResult = {smiles: smi, coreSmiles: core.originalSmiles, rGroupSmilesByNum: originalRgs};
+    const item: ChemEnumResult = {smiles: smi, coreSmiles: core.originalSmiles, rGroupSmilesByNum: rgSmiByNum};
```
drizhina, line 701: just curious, why?

packages/SequenceTranslator/src/polytool/pt-chem-enum-dialog.ts
```diff
@@ -458,6 +468,193 @@ async function openImportWizard(
...
+  return templatesPromise ??= (async () => {
+    let templates: RGroupTemplate[] = [];
+    try {
+      const text = await _package.files.readAsText('enumeration/r-group-templates.json');
+      templates = parseRGroupTemplates(text)
+        .map((t) => ({...t, items: t.items.filter((it) => isValidTemplateSmiles(it.smiles, rdkit))}))
+        .filter((t) => t.items.length > 0);
+    } catch { /* fall back to the built-in set below */ }
+    return templates.length > 0 ? templates : BUILTIN_R_GROUP_TEMPLATES;
+  })();
```
drizhina, line 587: lets make the following:
in files/enumeration folder, lets make another folder called r-group-templates.
place your template json there.
then in this function, use files api to list the folder, filter only jsons by extension and load all. this way, we will ship the default templates, but the users can also define their own and add to this folder. this is a really nice usability thing for clients. +, for the hardcoded file name of yours, lets sort the files so that templates from our file are last, and other templates from clients are first.

packages/SequenceTranslator/src/polytool/pt-chem-enum.ts
```diff
@@ -217,6 +217,146 @@ export function makeRGroup(
...
+  rGroupsByNum: Map<number, ChemEnumRGroup[]>,
+  smilesList: string[], targetN: number, mode: 'append' | 'replace', rdkit: RDModule,
+): number {
+  const made = smilesList.map((smi) => makeRGroup(smi, targetN, '', rdkit));
+  if (mode === 'replace') {
+    if (made.length > 0) rGroupsByNum.set(targetN, made);
+    else rGroupsByNum.delete(targetN); // replacing with nothing clears the slot
+  } else if (made.length > 0) {
+    const target = rGroupsByNum.get(targetN) ?? [];
+    target.push(...made);
```
drizhina: please do de-dupe here

drizhina (changes requested):
overall, looks good!
please address my comments
also, I don't see the third point from the ticket here - result table name and removing duplicates from result.

talagayev:
> overall, looks good! please address my comments also, I don't see the third point from the ticket here - result table name and removing duplicates from result.

Ah nice :) regarding the results table name and removing duplicates from result I split the PRs in the GROK subgroup tickets, so that one is the GROK-20222 one and the table name + removing duplicates is written GROK-20223 one, so I am doing that one in a separate branch so that it could be merged through a separate PR and the defaults one is then a separate one in GROK-20224, so thought to split them in the subtickets from the Markush Enumerator Ticket :)

packages/SequenceTranslator/src/polytool/pt-chem-enum.ts
talagayev: Ah here the main reason that I wanted to change it was connected with the case of the screenshot down below, that I thought that it would be confusing to have *:1 in the rest, while the Group that I want to add it to being R2 thus the options were to either code it that way, that by changing the R Group the number would change, so in R2 it would be *:2 etc. or remove it completly to not have the confusing case of *:1 displayed rests moving to R2 :)

Screenshot 2026-06-16 131944

talagayev: Alternative would be basically remapping during the R Group change, would probably be a better looking solution 🤔

packages/SequenceTranslator/src/polytool/pt-chem-enum-dialog.ts
talagayev: Removed that part

packages/SequenceTranslator/src/polytool/pt-chem-enum-dialog.ts
talagayev: Changed to icons with the template being both icon + name and download being and icon :)

packages/SequenceTranslator/src/polytool/pt-chem-enum-dialog.ts
talagayev: Done

packages/SequenceTranslator/src/polytool/pt-chem-enum-dialog.ts
talagayev: moved it there.

packages/SequenceTranslator/src/polytool/pt-chem-enum.ts
talagayev: Done

packages/SequenceTranslator/src/polytool/pt-chem-enum.ts
```diff
@@ -217,6 +217,158 @@ export function makeRGroup(
...
+
+/** Parses the templates JSON (an array of {@link RGroupTemplate}). Returns `[]` on anything
+ * malformed — callers fall back to {@link BUILTIN_R_GROUP_TEMPLATES}. This only enforces the
+ * shape; SMILES validity is checked separately by {@link invalidTemplateSmiles}. */
+export function parseRGroupTemplates(text: string): RGroupTemplate[] {
+  let raw: unknown;
+  try {
+    raw = JSON.parse(text);
+  } catch {
+    return [];
```
drizhina: generally not a good practice. at least do console.error

drizhina (approved):

packages/SequenceTranslator/src/polytool/pt-chem-enum.ts
talagayev: adjusted it

PR #3854, opened 2026-06-18
GROK-20223: SequenceTranslator: Markush enumeration - enum options

talagayev:
Changes made:
- Added option to remove duplicates in final output
- Added option to change the name of the output DF


packages/SequenceTranslator/src/polytool/pt-chem-enum.ts
```diff
@@ -755,3 +755,21 @@ export function enumerateSampleRaw(
...
 }
+
+/**
+ * First-occurrence keep-mask for `values`: `mask[i]` is `true` when `values[i]` is the
+ * first time that value is seen, `false` for a later duplicate. Nullish/empty entries are
+ * always kept — they stand for blank or unparseable rows that must not be collapsed together.
+ */
+export function uniqueKeepMask(values: ReadonlyArray<string | null | undefined>): boolean[] {
+  const seen = new Set<string>();
+  const mask = new Array<boolean>(values.length);
```
drizhina: why not create bitset here from start, and use bitset.setFast method. or bitset.init

drizhina: return bitset from here and you won't need to create it again above

talagayev: adjusted to return bitset :)

drizhina (approved):

PR #3859, opened 2026-06-19
Chem: Addition of Similarity Search filtering and selection option

talagayev:
Changes made:

- Added option of filtering and selecting molecules in Similarity search
- Adjusted edge case of similarity search restarting after every batch of substructure search results


packages/Chem/src/analysis/chem-search-base-viewer.ts
```diff
@@ -9,6 +9,8 @@ import {pickTextColorBasedOnBgColor} from '../utils/ui-utils';
 
 export const SIMILARITY = 'similarity';
 export const DIVERSITY = 'diversity';
+// Shared default cap (kept small — the diversity viewer's selection is O(n^2)). The similarity
+// viewer raises its own `limit` property max to MAX_LIMIT_SIMILARITY in its constructor.
 export const MAX_LIMIT = 50;
```
drizhina: didn't we want to change it to 200?

packages/Chem/src/analysis/chem-search-base-viewer.ts
```diff
@@ -65,10 +70,25 @@ export class ChemSearchBaseViewer extends DG.JsViewer {
...
+    // accumulate dead subs (and a second live onResetFilterRequest would toggle the filter twice).
+    this.subs = [];
   }
 
   async onTableAttached(): Promise<void> {
+    // Drop any prior subscriptions before re-subscribing (mirrors detach), so a re-attach WITHOUT a
+    // preceding detach can't accumulate duplicate subs — e.g. a second onResetFilterRequest that would
+    // make "Reset all filters" toggle the similarity filter twice.
+    this.subs.forEach((sub) => sub.unsubscribe());
+    this.subs = [];
```
drizhina: ?? where did this come from?

packages/Chem/src/analysis/chem-similarity-viewer.ts
```diff
@@ -33,7 +90,19 @@ export class ChemSimilarityViewer extends ChemSearchBaseViewer {
...
   constructor() {
     super(SIMILARITY);
-    this.cutoff = this.float('cutoff', 0.01, {min: 0, max: 1});
+    // Per-viewer property overrides: raise the limit cap, default row source to 'Filtered'. Done here (not
+    // onTableAttached) so a restored value isn't clamped against a stale max.
+    const limitProp = this.getProperty('limit');
+    if (limitProp) {
+      limitProp.max = MAX_LIMIT_SIMILARITY;
+      limitProp.defaultValue = DEFAULT_LIMIT_SIMILARITY; // 'Reset to default' / project default → 200
+    }
```
drizhina: well this is obviously not the way to do this. just have a getter on base class with default limit max and override it in this class. will work out of the box

packages/Chem/src/analysis/chem-similarity-viewer.ts
```diff
@@ -33,7 +90,19 @@ export class ChemSimilarityViewer extends ChemSearchBaseViewer {
...
+    this.limit = DEFAULT_LIMIT_SIMILARITY;
+    const rowSourceProp = this.getProperty('rowSource');
+    if (rowSourceProp)
+      rowSourceProp.defaultValue = RowSourceTypes.Filtered;
+    this.rowSource = RowSourceTypes.Filtered;
```
drizhina: same here

packages/Chem/src/analysis/chem-similarity-viewer.ts
```diff
@@ -94,10 +203,413 @@ export class ChemSimilarityViewer extends ChemSearchBaseViewer {
...
+    grid.classList.toggle('d4-current', isCurRow || isRef);
+    grid.classList.toggle('d4-selected', isSelected);
+    grid.style.boxShadow = isRef ? '0px 0px 1px var(--grey-6)' : '';
+    grid.style.backgroundColor = isSelected ?
+      ((isCurRow || isRef) ? '#d3f8bd' : '#f8f8df') : (isCurRow ? '#ddffd9' : '');
+  }
+
+
+  /** True when two masks are equal — compares packed word buffers directly (no clone), bailing early. */
+  private _sameMask(a: DG.BitSet | null, b: DG.BitSet | null): boolean {
```
drizhina: we reaaaly don't need methods on class that are one liners, and implementation is also not ideal. its one liner, bitsetA.xOr(bitsetB).trueCount === 0 does the same in one line and better :D

packages/Chem/src/analysis/chem-similarity-viewer.ts
```diff
@@ -94,10 +203,413 @@ export class ChemSimilarityViewer extends ChemSearchBaseViewer {
...
+    this._searchProgressSub?.unsubscribe();
+    this._searchProgressSub = null;
+    this._resetFilterState();
+    // If we were filtering, re-run the filter cycle (our listener is gone) so the table is restored
+    // instead of left filtered with no viewer or label to explain why.
+    if (wasFiltering && this.dataFrame)
+      this.dataFrame.rows.requestFilter();
+    super.detach();
+  }
+
```
drizhina, line 150: this is not the way I asked you too implement it. too much overcomplication, timers, too much code, only handing substructure filter specifically while other filters can be present only. 
all of the above needs to go. 
ontableattached is a tricky buisness, do not call it externally from anywhere.
all this can be fitter in few lines and much better. 

regardless of the rowsource, you recalculate the bitset only when your input param changes - i.e. the table or the template molecule or the threshold. you do this in debounced way. you do not recalculate the bitset when other filters change. you just use the original bitset to 'and' with filter. and then use it elsewhere.

you should only subscribe to onrowsfiltering event, without debounce, in ontableattached and in there, only if the filtering is enbled, AND the bitset with the filter. thats it. when things change in the input parameters or the bitset changes, you just request row filter, thats it. I don't understand why all of the above is needed, should be fitted in less then 100 lines.

drizhina (changes requested):

packages/Chem/src/analysis/chem-similarity-viewer.ts
```diff
@@ -109,30 +159,36 @@ export class ChemSimilarityViewer extends ChemSearchBaseViewer {
...
+            progressBar = DG.TaskBarProgressIndicator.create(`Similarity search running...`);
+            const res = await chemSimilarityBitset(moleculeColumn, this.targetMolecule,
+              this.distanceMetric as BitArrayMetrics, this.cutoff, this.fingerprint as Fingerprint);
+            if (!res) {
+              this.closeWithError(`Malformed molecule cannot be used for similarity search`, progressBar);
+              return;
+            }
+            this.simScores = res.scores;
+            this.simBitset = res.bitset;
+            this.computeKey = computeKey;
```
drizhina: I think this should be directly after the if statement, so that we do not go inside the if multiple times while waiting from async similarity bitset method

drizhina (commented):
Looks much cleaner!
only one minor comment

packages/Chem/src/analysis/chem-similarity-viewer.ts
talagayev: moved it after the if statement.

PR #3863, opened 2026-06-21
GROK-20224: SequenceTranslator: Markush enumeration - Defaults in plugin settings

talagayev:
Changes made:
- Addition of option to save package settings for a user group


packages/SequenceTranslator/src/polytool/pt-chem-enum-settings.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    w.root.appendChild(summary);
+    w.root.appendChild(ui.divH([editButton, clearButton], {style: {gap: '8px', marginTop: '6px'}}));
+  }
+
+  const monomersProp = propList.find((p) => p.name === monomersPathPropName);
+  if (monomersProp) {
+    const defaultTooltip = monomersProp.description ?? 'Path to additional monomer libraries';
+    let curTooltip = defaultTooltip;
+    const curValue: string = monomersProp.get(null) ?? monomersProp.defaultValue; // null is passed here because there is an getter override in the core
```
drizhina: probably in this case || is more correct rather then ??

packages/SequenceTranslator/src/polytool/pt-chem-enum-settings.ts
```diff
@@ -0,0 +1,149 @@
...
+    });
+    ui.tooltip.bind(clearButton, 'Remove the configured defaults; the app falls back to the shipped CSV files');
+
+    w.root.appendChild(summary);
+    w.root.appendChild(ui.divH([editButton, clearButton], {style: {gap: '8px', marginTop: '6px'}}));
+  }
+
+  const monomersProp = propList.find((p) => p.name === monomersPathPropName);
+  if (monomersProp) {
+    const defaultTooltip = monomersProp.description ?? 'Path to additional monomer libraries';
```
drizhina: same, || would be more correct

drizhina (commented):
Looks good. a super minor comment.

Can you also send over a small gif or screengrab of how it looks like?

talagayev:
> Looks good. a super minor comment.
> 
> Can you also send over a small gif or screengrab of how it looks like?

Ah yes can do :) 

Here is the GIF of an example
Markush_enumerator_package_settings

packages/SequenceTranslator/src/polytool/pt-chem-enum-settings.ts
talagayev: adjusted accordingly

PR #3868, opened 2026-06-24
GROK-18671: PowerGrid: Resolve form-referenced columns by their backing grid column instead of by data-column name in FormCellRenderer.makeScene

LaykDimon:
## GROK-18671: Forms cell renderer breaks when a referenced column is renamed via the grid header

**Bug.** Add a Designed-form (Smart Form) summary column referencing some columns (e.g. AGE/SEX/RACE), then rename one of those columns through the grid header. Every form cell turns into a red error X. PowerGrid's `FormCellRenderer.makeScene` (`public/packages/PowerGrid/src/forms/forms.ts`) resolves each referenced column's grid cell with `grid.cell(col.name, row)`, keyed on the *data* column name. A grid-header rename sets `GridColumn.customName` without renaming the underlying data column, so `GridColumnList.byName(dataName)` no longer matches; the resulting null `GridColumn` is dereferenced in `GridCell.gridCell` (d4 `grid_cell.dart:124`), throwing `NullError: method not found: 'column' on null`.

**Fix.** In `makeScene`, build a map from data-column name to the grid column's display name (scanning grid columns once, bounded to the referenced set), and resolve cells and field labels through that display name. This works for both rename paths and degrades gracefully: a referenced column with no grid column is now skipped instead of crashing. Field labels/tooltips also follow the grid display name, so the form reflects the rename.

The fix is scoped to the affected component rather than the core `GridColumn.name` setter: the data column is not renamed by a header rename, so there is nothing to sync, and making the setter emit `onColumnNameChanged` would misinform every other consumer of that event.

Crash site: `PowerGrid` / `src/forms/forms.ts` / `FormCellRenderer.makeScene`. Error pattern: `NullError: method not found: 'column' on null`.

**Tested.** `npm run build` exits 0 for PowerGrid (webpack compiled successfully). The renderer is canvas/grid-bound, so no unit test was added.

**Known limitation.** If a referenced column has no grid column at all (removed from the grid), that field is now omitted from the form rather than rendered — previously it crashed the whole form cell.

### DIFF
--- a/packages/PowerGrid/CHANGELOG.md
+++ b/packages/PowerGrid/CHANGELOG.md
@@ -7,6 +7,7 @@
 * Tags cell renderer: Fixed missing return in `getColor` that left first render of a new tag with an undefined color
 * Pie chart sparkline: Filter null columns in hit-test so hovering with stale/missing column names no longer throws
 * Forms cell renderer: Scoped per-cell scene via WeakMap keyed on grid (removes cross-grid contamination) and debounced tooltip shows (no more flicker)
+* GROK-18671: Forms cell renderer: Resolve referenced columns by their backing grid column so renaming a column via the grid header no longer breaks the form (red error cells); labels now follow the grid display name
 * Vlaaivis manager: Migrated drop handler to the new `doDrop(args)` signature in `ui.makeDroppable`
 
 ## 1.8.1 (2026-03-17)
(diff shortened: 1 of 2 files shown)


LaykDimon:
completely another feature, it should rather address the issue with form design

PR #3869, opened 2026-06-25
GROK-19013: Bio: Gate Bio/Peptide sequence viewers and tree viewers in the Add Viewer gallery by required column/tag, so they grey out on incompatible data

LaykDimon:
## GROK-19013: Add Viewer dialog: Bio/Peptide and tree viewers can be added to incompatible datasets

In the **Add Viewer** gallery, Bio/Peptide sequence viewers (WebLogo, VdRegions, and the Peptides viewers) and the tree viewers (Dendrogram, PhylocanvasGL) rendered as fully enabled/clickable cards even on datasets they cannot visualize (e.g. `System:DemoFiles/demog.csv`, which has no Macromolecule or Newick column). They should be greyed-out/disabled, the same way the Chem/Sequence Similarity & Diversity viewers already are after GROK-19005.

### Cause
`getJsViewers()` in `PowerPack/src/viewers-gallery.ts` only runs a compatibility `checkFunc` for viewers whose `friendlyName` is a key in a hardcoded per-package map. Viewers not covered fall through to the `else` branch where `isViewerEnabled` stays `true`, so their cards never receive the `disabled` class. The crash-site package in triage was `Bio`, but the enablement decision is owned entirely by the PowerPack gallery — there is no viewer-registration-level hook the gallery consults (`DG.Viewer.canVisualize` covers only core viewers).

### Fix
Following the reviewer-accepted GROK-19005 pattern (commit `da8c4c3e48`):
- Added `WebLogo`, `VdRegions`, and the Peptides viewers to `BIOCHEM_VIEWERS_TEST_DATA` (require a `Macromolecule` column) and added `'Peptides'` to that branch's package filter.
- Added a `TREE_VIEWERS_TEST_DATA` map for `Dendrogram` and `PhylocanvasGL`, enabled only when the table has a `Newick` column or a `.newick` tree tag.
- Threaded the new map into the disabled-card tooltip chain.

No existing branch is modified, so Chem/Charts/Curves and `showInGallery` behavior is unchanged.

### Testing
- `npm run build` in `packages/PowerPack` exits 0 (webpack compiled successfully).
- Verified against the reproduction recipe: on `demog.csv` every added `checkFunc` returns false, so the affected viewers receive the `disabled` class.

### Notes / scope
- The genuine root cause and only viable fix site is `PowerPack/src/viewers-gallery.ts`; `triage.required_packages` did not list PowerPack (a default platform plugin), so a reviewer should confirm scope.

See JIRA ticket GROK-19013 for reproduction evidence and screenshots.

### DIFF
--- a/packages/PowerPack/src/viewers-gallery.ts
+++ b/packages/PowerPack/src/viewers-gallery.ts
@@ -82,6 +82,37 @@ const BIOCHEM_VIEWERS_TEST_DATA: {[key: string] : ViewersTestData} = {
     tooltip: 'Chem Similarity Search viewer needs at least 1 sequence column'},
   'Sequence Diversity Search': {checkFunc: (table) => table.columns.bySemType(DG.SEMTYPE.MACROMOLECULE) !== null,
     tooltip: 'Chem Diversity Search viewer needs at least 1 sequence column'},
+  'WebLogo': {checkFunc: (table) => table.columns.bySemType(DG.SEMTYPE.MACROMOLECULE) !== null,
+    tooltip: 'WebLogo viewer needs at least 1 macromolecule column'},
+  'VdRegions': {checkFunc: (table) => table.columns.bySemType(DG.SEMTYPE.MACROMOLECULE) !== null,
+    tooltip: 'VdRegions viewer needs at least 1 macromolecule column'},
+  'Sequence Variability Map': {checkFunc: (table) => table.columns.bySemType(DG.SEMTYPE.MACROMOLECULE) !== null &&
+    [...table.columns.numerical].length >= 1,
+    tooltip: 'Sequence Variability Map viewer needs a macromolecule and a numeric column'},
+  'Most Potent Residues': {checkFunc: (table) => table.columns.bySemType(DG.SEMTYPE.MACROMOLECULE) !== null &&
+    [...table.columns.numerical].length >= 1,
+    tooltip: 'Most Potent Residues viewer needs a macromolecule and a numeric column'},
+  'Sequence Mutation Cliffs': {checkFunc: (table) => table.columns.bySemType(DG.SEMTYPE.MACROMOLECULE) !== null &&
+    [...table.columns.numerical].length >= 1,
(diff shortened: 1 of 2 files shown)


packages/PowerPack/src/viewers-gallery.ts
```diff
@@ -82,6 +82,29 @@ const BIOCHEM_VIEWERS_TEST_DATA: {[key: string] : ViewersTestData} = {
     tooltip: 'Chem Similarity Search viewer needs at least 1 sequence column'},
   'Sequence Diversity Search': {checkFunc: (table) => table.columns.bySemType(DG.SEMTYPE.MACROMOLECULE) !== null,
     tooltip: 'Chem Diversity Search viewer needs at least 1 sequence column'},
+  'WebLogo': {checkFunc: (table) => table.columns.bySemType(DG.SEMTYPE.MACROMOLECULE) !== null,
+    tooltip: 'WebLogo viewer needs at least 1 macromolecule column'},
+  'VdRegions': {checkFunc: (table) => table.columns.bySemType(DG.SEMTYPE.MACROMOLECULE) !== null,
+    tooltip: 'VdRegions viewer needs at least 1 macromolecule column'},
+  'Sequence Variability Map': {checkFunc: (table) => table.columns.bySemType(DG.SEMTYPE.MACROMOLECULE) !== null,
```
drizhina: one macromolecule and one numeric column needs to be present as well.
this is true also for:
'Most Potent Residues'
'Sequence Mutation Cliffs'
'Sequence Position Statistics'

For 
'Active peptide selection': requires sequences, 1 numeric column + 1 numeric / categorical column
'Logo Summary Table' requires 1 macromolecule + 1 numeric + 1 numeric / categorical

drizhina (changes requested):

packages/PowerPack/src/viewers-gallery.ts
LaykDimon: addressed it now

PR #3870, opened 2026-06-25
GROK-20246: Markush Enum: UI redesign

talagayev:
Modification of Markush Enumerator UI


talagayev:
Here is a GIF currently displaying the UI and interactions with it

markush_redesign_UI

PR #3872, opened 2026-06-25
[ticket] GROK-19652: Chem: Set V2000 counts-line chiral flag in _convertMolNotation when the structure has defined absolute stereo

LaykDimon:
## GROK-19652: Chem: stereochemistry changes when opening the sketcher from a molecule preview

RDKit's get_molblock() emits a V2000 chiral flag of 0 even for defined absolute stereocenters, so OCL/Ketcher load them as racemate and drop @/@@ on export. Fixed at the shared producer _convertMolNotation by setting the counts-line chiral flag to 1 for structures with a CIP R/S stereocenter. npm run build passes; added a regression test.

### DIFF
--- a/packages/Chem/CHANGELOG.md
+++ b/packages/Chem/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v.next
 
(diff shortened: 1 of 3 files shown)


drizhina:
this is very sketchy and I would not do it. for enhanced stereo we should use v3000, and it has already been improved in new ketcher

LaykDimon:
thanks! will reuse it to tune

PR #3880, opened 2026-06-30
GROK-20247: Chemical enumeration: Redesign config configuration

talagayev:
Changes made:

- Modified UI of Reaction Enumeration


drizhina:
Overall good, but:
too many inline css, use css files as much as possible. cssText is quite bad.
also, 
you clearly intended a light blue hover (the #e9f3fb fallback). But --blue-1 is defined — it's #2083D5, a saturated dark blue.
The fallback never fires. Chip text is --grey-6 (#4D5261, dark) → dark text on dark-blue background on hover = unreadable.

you also have 
chip.style.borderColor = err ? 'var(--red-2)' : 'var(--grey-3)';
but --red-2 does not exist :D :D probably a claude artifact. 
The error-state border resolves to an invalid value and the border doesn't change. The error background (--red-1) works, so it half-renders. Use --red-3 (the defined error/failure red, also used for validationDiv).

drizhina (changes requested):
left comment on PR

talagayev:
> Overall good, but: too many inline css, use css files as much as possible. cssText is quite bad. also, you clearly intended a light blue hover (the #e9f3fb fallback). But --blue-1 is defined — it's #2083D5, a saturated dark blue. The fallback never fires. Chip text is --grey-6 (#4D5261, dark) → dark text on dark-blue background on hover = unreadable.
> 
> you also have chip.style.borderColor = err ? 'var(--red-2)' : 'var(--grey-3)'; but --red-2 does not exist :D :D probably a claude artifact. The error-state border resolves to an invalid value and the border doesn't change. The error background (--red-1) works, so it half-renders. Use --red-3 (the defined error/failure red, also used for validationDiv).

Nice nice, will adress those :)

packages/Chem/css/chem.css
```diff
@@ -1275,3 +1253,113 @@
...
+
+.chem-enumerator .d4-accordion-pane-content .ui-input-editor {
+  flex-grow: 1;
+  min-width: 0;
+}
+
+/* The platform lays out views as: .d4-ribbon ~ view.root (.chem-enumerator-view).
+   These :has(~ .chem-enumerator-view) rules scope ribbon overrides to this view only. */
+.d4-ribbon:has(~ .chem-enumerator-view) {
+  background: var(--white);
```
drizhina: why? its already white

packages/Chem/css/chem.css
```diff
@@ -1275,3 +1253,113 @@
...
+}
+
+.d4-ribbon:has(~ .chem-enumerator-view) .d4-ribbon-item:hover {
+  background: transparent !important;
+}
+
+.d4-ribbon:has(~ .chem-enumerator-view) .d4-ribbon-item > :hover:not(button):not(.ui-btn):not(.chem-enum-chip) {
+  background: transparent !important;
```
drizhina: not sure why either

packages/Chem/css/chem.css
```diff
@@ -1275,3 +1253,113 @@
...
+}
+
+.d4-ribbon:has(~ .chem-enumerator-view) .d4-ribbon-separator {
+  display: none;
```
drizhina: again, why? you have the ability to control the ribbon, and based on the array you pass, separator will be drawn like that

drizhina: plus, the 'has' thing is relatively new. It does not work in some browsers

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
```diff
@@ -1097,42 +1080,67 @@ export async function buildEnumeratorView(): Promise<DG.ViewBase> {
...
-    headerRow,
     mainRow,
     validationDiv,
-    actionBar,
-  ], {style: {padding: '16px', height: '100%', boxSizing: 'border-box', overflow: 'hidden'}});
+  ], {style: {padding: '0 0 0 16px', height: '100%', boxSizing: 'border-box', overflow: 'hidden'},
+    classes: 'chem-enumerator'});
 
+  view.root.classList.add('chem-enumerator-view');
+  view.setRibbonPanels([[appInfoIcon], [runGroup], [chipReactions], [ribbonArrow], [chipBbs], [chipCombine], [cfgEstEl]]);
```
drizhina: this is why you get all the separators. why don't you just put all in one array?

packages/Chem/css/chem.css
talagayev: was worried about the grey highlighting in the background in the ribbon, so was cosmetic, but didn't really help. Fixed it :)

packages/Chem/css/chem.css
talagayev: Same case, fixed

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
talagayev: put them all in one array

packages/Chem/css/chem.css
talagayev: True that, adjusted

PR #3889, opened 2026-07-05
GROK-20248: Reaction enum: Allow subsets of reactions / ingredients / bbs for each step

talagayev:


packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
```diff
@@ -482,72 +495,145 @@ export async function buildEnumeratorView(): Promise<DG.ViewBase> {
...
+  // manually per column instead of asking the platform to auto-detect.
+  function mountDf(host: HTMLElement, df: DG.DataFrame, withFilters: boolean): void {
+    closeLiveFilterView(host);
+    host.innerHTML = '';
     const grid = DG.Viewer.grid(df);
     grid.props.rowHeight = GRID_ROW_HEIGHT;
-    grid.root.style.width = '100%';
-    grid.root.style.height = '100%';
-    host.appendChild(grid.root);
+    grid.root.style.cssText += ';width:100%;height:100%';
```
drizhina: I did ask you not to use csstext on many previous prs

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
```diff
@@ -482,72 +495,145 @@ export async function buildEnumeratorView(): Promise<DG.ViewBase> {
...
+      const filterGroup = tv.getFiltersGroup({createDefaultFilters: false});
+      for (const col of df.columns.toList()) {
+        if (FILTER_CRASH_SEMTYPES.includes(col.semType)) continue;
+        const type = col.isNumerical ? DG.FILTER_TYPE.HISTOGRAM :
+          col.semType === 'Molecule' ? DG.FILTER_TYPE.SUBSTRUCTURE : DG.FILTER_TYPE.CATEGORICAL;
+        try {filterGroup.updateOrAdd({type, column: col.name});} catch (e) {
+          console.warn(`Could not add a default filter for column "${col.name}":`, e);
+        }
+      }
+      filterGroup.root.style.cssText += ';width:100%;height:100%;overflow:auto';
```
drizhina: same here

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
```diff
@@ -482,72 +495,145 @@ export async function buildEnumeratorView(): Promise<DG.ViewBase> {
...
+  function mountDf(host: HTMLElement, df: DG.DataFrame, withFilters: boolean): void {
+    closeLiveFilterView(host);
+    host.innerHTML = '';
     const grid = DG.Viewer.grid(df);
     grid.props.rowHeight = GRID_ROW_HEIGHT;
-    grid.root.style.width = '100%';
-    grid.root.style.height = '100%';
-    host.appendChild(grid.root);
+    grid.root.style.cssText += ';width:100%;height:100%';
+    if (!withFilters) {
```
drizhina: why do you handle filters customly? just having a tableview in there and adding filters to it natively instead of having grid and having to manually do all that with splitH

drizhina: better yet, not even sure why do you need a tableview at all and all that handing.
you can either use DG.Viewer.filters(df, {columnNames})
or dataframe.plot.filter({columnNames})

drizhina (changes requested):
renderGrid() unconditionally rebuilds a Dart-side grid on every step-tab switch, every depth/breadth toggle (dataCtls.forEach((c) => c.render()) remounts all three panels to refresh a dot), and twice per reagents change; one "Subset by selection" click now triggers a double rebuild (triple for reagents) because the old if (state.suppress) return; guard on programmatic swaps was dropped. Discarded grids are never close()d — the PR's own comment explains why innerHTML = '' leaks for TableViews and carefully tracks those, but applies the identical untracked pattern to grids. Mode/reagent changes need only renderBar()+updateDots(); grids should be cached per pane and closed on replacement.

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
talagayev: removed cssText

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
talagayev: adressed

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
talagayev: changed to `DG.Viewer.filters`

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
```diff
@@ -1113,33 +1726,51 @@ export async function buildEnumeratorView(): Promise<DG.ViewBase> {
...
+  function applyRibbonFixup(attempt = 0): void {
+    const el = runBtn.closest<HTMLElement>('.d4-ribbon-item');
+    if (el) {
+      document.querySelectorAll<HTMLElement>('.d4-ribbon-group, .d4-ribbon-item').forEach((g) => {
+        g.style.background = 'transparent';
+        g.style.boxShadow = 'none';
+        g.style.border = 'none';
+      });
+    }
+    if (attempt < 200) setTimeout(() => applyRibbonFixup(attempt + 1), 50);
```
drizhina: what is this exactly? if I open this view, and switch to any other view, it will permanently restyle the ribbon panel of the other views.

drizhina (changes requested):

drizhina:
I would suggest it to have a small header 'Reaction Enumeration' and a one line subheader here, otherwise looks odd
image

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
```diff
@@ -478,86 +552,188 @@ export async function buildEnumeratorView(): Promise<DG.ViewBase> {
...
+  // explicitly — track the live ones per host and close them before mounting a replacement.
+  const mountedViewers = new Map<HTMLElement, DG.Viewer[]>();
+  function closeMountedViewers(host: HTMLElement): void {
+    const prev = mountedViewers.get(host);
+    if (!prev) return;
+    mountedViewers.delete(host);
+    for (const v of prev) {
+      // A viewer mid-Dart-side-construction throws TypeError on close — expected on initial
+      // load, not actionable.
+      try {v.close();} catch (e) {
```
drizhina: please pay attention to such stuff. close is not going to do anything absolutely, not even detach. so what you should do is do v.detach() and then v.root.remove

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
```diff
@@ -478,86 +552,188 @@ export async function buildEnumeratorView(): Promise<DG.ViewBase> {
...
+        if (!(e instanceof TypeError)) console.warn('Could not close previous viewer:', e);
+      }
     }
+  }
+  // Neither of the above runs on VIEW close — only on remount/no-table. Close every still-mounted
+  // viewer when the view itself tears down, or their Dart-side resources leak for the session.
+  view.subs.push(new Subscription(() => {
```
drizhina: what is this exactly?

drizhina: have you checked this at least? :D

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
```diff
@@ -478,86 +552,188 @@ export async function buildEnumeratorView(): Promise<DG.ViewBase> {
...
+  // when both are present, since it's the more deliberate action.
+  function cloneSubsetByRows(df: DG.DataFrame, emptyMsg: string): DG.DataFrame | null {
+    const sel = df.selection;
+    const mask = sel.trueCount === 0 ? df.filter : sel;
+    if (mask.trueCount === 0) {
+      grok.shell.info(emptyMsg);
+      return null;
+    }
+    // Also covers the trivial "filter matches every row" case — nothing left to narrow either way.
+    if (mask.trueCount === df.rowCount) {
```
drizhina: fyi, there is .allTrue getter

drizhina (changes requested):

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
talagayev: adjusted to use v.detach() and then v.root.remove

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
talagayev: adressed it 🙈

packages/Chem/css/chem.css
```diff
@@ -1298,12 +1298,51 @@
   min-width: 0;
 }
 
+/* The platform marks a null column-selector `.d4-invalid` regardless of `nullable` — not an error
+   here. Its own rule uses `!important`, so the override needs one too. */
+.chem-enum-optional-col .d4-invalid {
```
drizhina: this is not correct of course. its not the issue with nullable at all. if you hover over it you will see that its complaining because there is no table, not because its empty. nullable is true by default. so what you should do is add a special class only in case if there is no table, and in such case, just disable that input and have tooltip to first choose a table.

packages/Chem/css/chem.css
```diff
@@ -1341,21 +1398,45 @@
...
+.chem-enum-ribbon-scroll {
+  scrollbar-width: thin;
+}
+
+.chem-enum-ribbon-scroll::-webkit-scrollbar {
+  height: 4px;
+}
+
+.chem-enum-ribbon-scroll::-webkit-scrollbar-thumb {
+  background: var(--grey-3);
```
drizhina: where did this come from? we never use this color for scrollbar. look at the image here. 
also, the whole section is scrollable, and its scrolling the main enumerate button. this is of course not good, only the div with steps should be scrollable.

Image

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
drizhina: you didn't address this comment. this is a very major one and we can't do stuff like this.

packages/Chem/src/widgets/chem-substructure-filter.ts
```diff
@@ -679,12 +682,23 @@ export class SubstructureFilter extends DG.Filter {
...
       false, false, this.searchType, this.similarityCutOff, this.fp);
   }
 
+  // `this.column` can end up null here even after attach() ran, if applyState() fired with
+  // `this.dataFrame` not yet set (a mount-timing race — live-observed: sketching a query silently
+  // no-oped instead of filtering, with `this.column!.temp` throwing under the hood). Re-resolve by
+  // name (or by semtype as a last resort) instead of trusting the cached, possibly-stale reference.
+  private resolveColumn(): DG.Column | null {
+    this.column ??= this.dataFrame && this.columnName ? this.dataFrame.col(this.columnName) : null;
+    this.column ??= this.dataFrame?.columns.bySemType(DG.SEMTYPE.MOLECULE) ?? null;
```
drizhina: this is wrong and reallllly bad sideeffect. we should not do this. I mean the last line, with this.dataFrame?.columns.bySemType(DG.SEMTYPE.MOLECULE) ?? null; it can go to another column.

drizhina (changes requested):

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
```diff
@@ -760,65 +1089,202 @@ export async function buildEnumeratorView(): Promise<DG.ViewBase> {
...
+  const attachChangedDot = (pane: DG.AccordionPane, tooltip: string): HTMLElement => {
+    const dot = mkChangedDot(tooltip);
+    dot.style.marginLeft = '6px';
+    pane.root.querySelector('.d4-accordion-pane-header')?.appendChild(dot);
+    return dot;
+  };
+  // A form measured while too narrow gets permanently marked .ui-form-condensed by the platform,
+  // which applies its own compact styling. Strip the class back off instead of mirroring what it
+  // changes — stays correct even if the platform's condensed styling changes.
+  const stripCondensedClass = (form: HTMLElement): MutationObserver => {
```
drizhina: this is straight fighting the platform. instead, just when creating the form, go through the labels of inputs, and set their max width to something reasonable. they are getting condensed because of long input names

drizhina: I see such stuff in many places. don't fight with platform, if there are such stuff, report

drizhina (changes requested):
"Use all" restores the wrong table — restoreFullTable/onTableChanged. state.original is only ever set inside subsetBySelection; swapping the table input to a different file doesn't touch it. So: load templates A → subset → swap input to file B → click "Use all" → the app silently replaces B with a clone of A. The code comment admits this trade-off, but this isn't an obscure edge case — "pick a file, change my mind, hit Use all" is a normal flow, and silent data substitution in a chemistry tool is how someone enumerates against the wrong library without noticing. The underlying cause is name-string identity (df.name !== state.prev?.name) standing in for real identity — table names aren't unique in a workspace. Recommend clearing state.original/prev/freshClone in onTableChanged when the incoming table matches neither prev nor freshClone, rather than never.

packages/Chem/src/widgets/chem-substructure-filter.ts
talagayev: fixed it to not guess and randomly grab a column

packages/Chem/css/chem.css
talagayev: modified the ribbon to allow for scrolling while restricting it to the part containing the chips to keep the remainder of the ribbon static, while making the ribbon not visible to have it in the same visual style like the remainder of the ribbon

packages/Chem/css/chem.css
talagayev: adjusted it through using  the `InputBase.enabled` here

packages/Chem/src/utils/reaction-enumeration/enumerator-app.ts
talagayev: adjusted the width instead of stripping the class. tried just modifying width first, but that didn't fully fix it and I saw the pane condense/uncondense depending on where the splitter was (condensed if the splitter was too far to the left, thus the left panel having less space), since the bottleneck for Product filters apparently is the input side since the `only these atoms allowed` filed needs 200px, so a narrower label alone doesn't stop the form crossing the condense threshold and I ended up sizing the whole form up front instead for it to work reliably regardless of splitter position.

drizhina:
1. Mounted viewers are no longer cleaned up on view close — regression from the previous revision

The earlier version had:


view.subs.push(new Subscription(() => {
  for (const host of mountedViewers.keys()) closeMountedViewers(host);
}));
That block is gone. closeMountedViewers is now only called from mountDf (:725), buildStepTabs (:1641), renderGrid's empty branch (:1772), and showInPreview (:2023) — all remount paths. Nothing runs on teardown.

mountedViewers holds a Grid (and a Filters viewer when the funnel is on) per gridHost, and there is one gridHost per round tab per panel — up to 3 × 11 = 33 hosts plus the preview grid. Close the view and every one of those Dart-side viewers leaks for the session. The file's own comment at :706 states exactly why this matters. Restore the teardown subscription.

2. Two enumerations can run concurrently

runWithUi disables only runBtn (:982) and re-enables it in finally (:991). previewEnumerateBtn (:949) calls the same runWithUi(runEnumeration) but its disabled state is owned solely by refreshValidation() (:641) — nothing touches it during a run.

So: click Enumerate in the ribbon, switch to the Preview pane, click Enumerate there. The second run sets cancelled = false, clobbering the first run's cancel flag; both write progressLabel; both call grok.shell.addTableView; the first run's finally clears the progress label and Cancel button while the second is still running. Cancel is effectively broken for the duration.

Add a single running guard in runWithUi and drive both buttons from it.

3. runWithUi's finally re-enables Run regardless of validity

Still present (:991). runBtn.disabled = false overrides whatever refreshValidation() decided — if the config went invalid during the run (a table closed, a column cleared), Run stays clickable until the next input change. Call refreshValidation() instead of assigning directly. Same for previewEnumerateBtn once #2 is fixed.

4. Preview silently ignores per-round subsets — and the disclaimer that said so was removed

refreshPreview calls enumerate({rdkit, config: previewConfig, ...inputs, isCancelled}) (:2093) with no perRoundOverrides. That was acceptable when the header read "per-step subsets are not applied here" — but that sentence was cut, and the header is now just 'Quick preview of a small product sample.' (:2014).

Worse, the same Preview pane now displays the per-round subsets: renderPreviewRecap (:1289–1301) prints Round r: N of M (custom subset) from buildPerRoundOverrides(config). So the panel tells the user round 2 uses 12 of 40 building blocks, then previews a run that used all 40. Either pass perRoundOverrides through, or put the disclaimer back.

5. pickFile() never resolves if the user cancels

Unchanged (:129–143). input.onchange only fires on selection; cancelling the OS dialog leaves the promise pending forever and orphans a hidden <input> in document.body — one per cancel. loadYamlBtn's await pickFile(...) never returns. Needs a cancel/focus fallback and unconditional cleanup.

7. The function is now ~1,950 lines, and correctness depends on nothing firing between declarations

buildEnumeratorView grew by 430 lines. The concrete cost, not just an aesthetic one: refreshCfgRibbon (:1372) dereferences templatesCtl, bbsCtl, reagentsCtl (:1396–1398) and tabs/strategyPane/previewPane (:1401–1402) — all consts declared 450–770 lines later. wireValidation registers onChanged → refreshValidation → refreshCfgRibbon at :902.

So any input onChanged firing between :902 and :1838 is a ReferenceError that kills view construction. accordion.addPane('Reactions', factory, true, …) at :1123 runs its factory synchronously (expanded = true), building ui.form([templatesInput, smartsColInput, blockingColInput, rxnNameColInput]). DG inputs can fire onValueChanged during construction — that's a known behavior in this codebase. This is the same landmine that already bit with tabs (documented at :1456 and :2204); it now exists one layer down and is undocumented.

Similarly syncConfigToQuickInputs (:542) references wireValidation (:893), combinationLimitsForm (:1243) and productFilterForm (:1244) — safe only because it's never called before the YAML button is clickable.

This is not fixable by another comment. Extract EnumeratorConfigForm, DataPanel (as a class), StrategySummary, and PreviewPanel into their own modules with explicit constructor wiring.

8. ~70 lines re-implementing the platform's form-layout algorithm by copying its constants

measureLabelWidths (:1203), platformInputMinWidth (:1233), sharedEditorWidth = 150 (:1227), +40 buffer (:1267), plus 9 !important CSS declarations pinning label/editor columns. The comments are explicit that these mirror js-api ui.ts's getInputsMinWidths / handleFormResize and the platform's .ui-form-condensed thresholds. Any change to those platform values silently mislays this pane out, with no test and no error.

This is the largest piece of duct tape in the current revision. The real fix is a ui.form option for a shared label column — worth a js-api ticket. Also, flushIndent (:1154) reads getBoundingClientRect() and writes margin-left from inside a ResizeObserver callback; the author avoided the obvious feedback loop by observing an ancestor, but it still forces synchronous layout on every resize.


9. FILTER_REMOUNT_SETTLE_MS = 200 — still an unguarded, uncancellable race

deferredFilterReset (:59) and withPreservedFilters (:916) schedule bare setTimeouts that aren't tracked. Close the view within 200 ms and they fire against a possibly-closed DataFrame. The 200 ms is a guess — on a slow machine the platform's async reapply lands after the fix and the bug returns. If the Filters viewer genuinely reapplies stale categorical state post-construction, that's a core bug that needs a ticket; at minimum keep the handles and clear them in a view.subs teardown.

10. No debounce on the rounds input

— every keystroke runs onRoundsChanged() on all three panels → buildStepTabs() → close all viewers, wipe the DOM, rebuild a TabControl, remount Grid + Filters. Typing 10 over 5 does that twice, three panels each. The code already documents the transient-1 problem at :1816–1818; debouncing solves both that and the flicker.



ALSO


Load YAML… silently discards almost every setting — the guard it relies on is dead code

and 

 buildEnumeratorView is ~1,550 lines in one function....


and 


Comments are a debugging diary, not documentation. Roughly 250 lines of the new file are narration of past attempts: "live-verified", "confirmed live", "two earlier attempts … were both live-verified to occasionally wipe", "a real double DOM rebuild per review feedback". I really don't want these in the codebase....


please reply for each point when you address them. try to follow code flow, review it yourself and with claude as well

talagayev:
> 1. Mounted viewers are no longer cleaned up on view close — regression from the previous revision
> 
> The earlier version had:
> 
> view.subs.push(new Subscription(() => { for (const host of mountedViewers.keys()) closeMountedViewers(host); })); That block is gone. closeMountedViewers is now only called from mountDf (:725), buildStepTabs (:1641), renderGrid's empty branch (:1772), and showInPreview (:2023) — all remount paths. Nothing runs on teardown.
> 
> mountedViewers holds a Grid (and a Filters viewer when the funnel is on) per gridHost, and there is one gridHost per round tab per panel — up to 3 × 11 = 33 hosts plus the preview grid. Close the view and every one of those Dart-side viewers leaks for the session. The file's own comment at :706 states exactly why this matters. Restore the teardown subscription.
> 
> 2. Two enumerations can run concurrently
> 
> runWithUi disables only runBtn (:982) and re-enables it in finally (:991). previewEnumerateBtn (:949) calls the same runWithUi(runEnumeration) but its disabled state is owned solely by refreshValidation() (:641) — nothing touches it during a run.
> 
> So: click Enumerate in the ribbon, switch to the Preview pane, click Enumerate there. The second run sets cancelled = false, clobbering the first run's cancel flag; both write progressLabel; both call grok.shell.addTableView; the first run's finally clears the progress label and Cancel button while the second is still running. Cancel is effectively broken for the duration.
> 
> Add a single running guard in runWithUi and drive both buttons from it.
> 
> 3. runWithUi's finally re-enables Run regardless of validity
> 
> Still present (:991). runBtn.disabled = false overrides whatever refreshValidation() decided — if the config went invalid during the run (a table closed, a column cleared), Run stays clickable until the next input change. Call refreshValidation() instead of assigning directly. Same for previewEnumerateBtn once #2 is fixed.
> 
> 4. Preview silently ignores per-round subsets — and the disclaimer that said so was removed
> 
> refreshPreview calls enumerate({rdkit, config: previewConfig, ...inputs, isCancelled}) (:2093) with no perRoundOverrides. That was acceptable when the header read "per-step subsets are not applied here" — but that sentence was cut, and the header is now just 'Quick preview of a small product sample.' (:2014).
> 
> Worse, the same Preview pane now displays the per-round subsets: renderPreviewRecap (:1289–1301) prints Round r: N of M (custom subset) from buildPerRoundOverrides(config). So the panel tells the user round 2 uses 12 of 40 building blocks, then previews a run that used all 40. Either pass perRoundOverrides through, or put the disclaimer back.
> 
> 5. pickFile() never resolves if the user cancels
> 
> Unchanged (:129–143). input.onchange only fires on selection; cancelling the OS dialog leaves the promise pending forever and orphans a hidden in document.body — one per cancel. loadYamlBtn's await pickFile(...) never returns. Needs a cancel/focus fallback and unconditional cleanup.
> 
> 7. The function is now ~1,950 lines, and correctness depends on nothing firing between declarations
> 
> buildEnumeratorView grew by 430 lines. The concrete cost, not just an aesthetic one: refreshCfgRibbon (:1372) dereferences templatesCtl, bbsCtl, reagentsCtl (:1396–1398) and tabs/strategyPane/previewPane (:1401–1402) — all consts declared 450–770 lines later. wireValidation registers onChanged → refreshValidation → refreshCfgRibbon at :902.
> 
> So any input onChanged firing between :902 and :1838 is a ReferenceError that kills view construction. accordion.addPane('Reactions', factory, true, …) at :1123 runs its factory synchronously (expanded = true), building ui.form([templatesInput, smartsColInput, blockingColInput, rxnNameColInput]). DG inputs can fire onValueChanged during construction — that's a known behavior in this codebase. This is the same landmine that already bit with tabs (documented at :1456 and :2204); it now exists one layer down and is undocumented.
> 
> Similarly syncConfigToQuickInputs (:542) references wireValidation (:893), combinationLimitsForm (:1243) and productFilterForm (:1244) — safe only because it's never called before the YAML button is clickable.
> 
> This is not fixable by another comment. Extract EnumeratorConfigForm, DataPanel (as a class), StrategySummary, and PreviewPanel into their own modules with explicit constructor wiring.
> 
> 8. ~70 lines re-implementing the platform's form-layout algorithm by copying its constants
> 
> measureLabelWidths (:1203), platformInputMinWidth (:1233), sharedEditorWidth = 150 (:1227), +40 buffer (:1267), plus 9 !important CSS declarations pinning label/editor columns. The comments are explicit that these mirror js-api ui.ts's getInputsMinWidths / handleFormResize and the platform's .ui-form-condensed thresholds. Any change to those platform values silently mislays this pane out, with no test and no error.
> 
> This is the largest piece of duct tape in the current revision. The real fix is a ui.form option for a shared label column — worth a js-api ticket. Also, flushIndent (:1154) reads getBoundingClientRect() and writes margin-left from inside a ResizeObserver callback; the author avoided the obvious feedback loop by observing an ancestor, but it still forces synchronous layout on every resize.
> 
> 9. FILTER_REMOUNT_SETTLE_MS = 200 — still an unguarded, uncancellable race
> 
> deferredFilterReset (:59) and withPreservedFilters (:916) schedule bare setTimeouts that aren't tracked. Close the view within 200 ms and they fire against a possibly-closed DataFrame. The 200 ms is a guess — on a slow machine the platform's async reapply lands after the fix and the bug returns. If the Filters viewer genuinely reapplies stale categorical state post-construction, that's a core bug that needs a ticket; at minimum keep the handles and clear them in a view.subs teardown.
> 
> 10. No debounce on the rounds input
> 
> — every keystroke runs onRoundsChanged() on all three panels → buildStepTabs() → close all viewers, wipe the DOM, rebuild a TabControl, remount Grid + Filters. Typing 10 over 5 does that twice, three panels each. The code already documents the transient-1 problem at :1816–1818; debouncing solves both that and the flicker.
> 
> ALSO
> 
> Load YAML… silently discards almost every setting — the guard it relies on is dead code
> 
> and
> 
> buildEnumeratorView is ~1,550 lines in one function....
> 
> and
> 
> Comments are a debugging diary, not documentation. Roughly 250 lines of the new file are narration of past attempts: "live-verified", "confirmed live", "two earlier attempts … were both live-verified to occasionally wipe", "a real double DOM rebuild per review feedback". I really don't want these in the codebase....
> 
> please reply for each point when you address them. try to follow code flow, review it yourself and with claude as well

1. Fixed to make the Mounted viewers cleaunup on view close

2. runWithUi now holds a single running guard checked before either button can start a run; both the ribbon Enumerate and the Preview-pane Enumerate button route through the same `runWithUi `call, so a second run can't start while one's in flight

3. Adjusted finally now calls `this.deps.refreshValidation()` to re-checks current validity rather than blindly re-enabling.

4. The `PreviewPanel.refresh()` now passes `perRoundOverrides` into `enumerate()`, so the actual preview run matches what the recap card displays.

5. Adjusted the code for `pickFile()` to recognize if the user cancels

7. Split the code into `MountedViewerRegistry`, `StrategySummary`, `PreviewPanel`, `DataPanel `, `RunControls` and `EnumeratorConfigForm `into their own modules.

8. modified to use constant widths on the labels

9.  `deferredFilterReset `and `withPreservedFilters `schedule their timers through one `scheduleTimer()` helper that tracks the id in a `pendingTimers `set, which the view-close handler clears alongside the viewers.

10. Rounds-changed rebuild is now debounced with `ROUNDS_INPUT_DEBOUNCE_MS`

For the YAML runs now `validateShape()` to check that everything is correct

For the comments went through to clean them up and reviewed the code both myself and CC

PR #3919, opened 2026-07-17
CVMTests: fix Grok Column list script — wrap list in Named()

alex-aprm:
## What

Fixes the failing **CVM Tests → Scripts: Grok scripts → Column list** test.

`scripts/grok/column_list.grok` called `DeleteColumns(df, [cols.first])`. `DeleteColumns`'s second parameter is a column-filter predicate (`ColFilterCall`). The parser only coerces a **parse-time** list literal (e.g. `DeleteColumns("cars", ["make","model"])`) into a `Named(...)` predicate — a **runtime** list value like `[cols.first]` reaches `deleteColumnsImpl` as a raw `List` and throws:

```
NoSuchMethodError: Class 'List' has no instance method 'makePredicate'
```

Wrapping the list in `Named([cols.first])` builds the `NamedColumnPredicate` explicitly, deleting the first column as intended (matching the Python/R/NodeJS/JS/Octave versions).

## Verification

- Reproduced the throw and confirmed the fix locally against the real Dart evaluation path (`Script.fromCode` → `GrokScriptHandler.run` → `Func.runScript` → `DeleteColumns`), result `(date, name)`.
- Ran the actual test on **dev**: `Scripts: Grok scripts / Column list` — **1 passed**.


alex-aprm:
Committed directly to master as 24731a3170. Closing this PR.

PR #3929, opened 2026-07-21
[ticket] GROK-20419: Tutorials: Point the Embedded Viewers final-step hint at the inline Trellis props host instead of a dead selector

LaykDimon:
The Embedded Viewers tutorial's final step showed no highlight because its hint used `.grok-font-icon-settings.d4-viewer-icon`, matching no element (those classes live on unrelated elements), so Tutorial.action's null-guard skipped hint placement. The inner scatter plot's Color is edited inline in the Trellis properties host (`.d4-trellis-plot-props-host`); the fix points the hint there and updates the stale gear-icon instruction text. Verified with `npm run build` (exit 0).

---
**Diff:** +4/-3 · 2 files · full analysis: [GROK-20419](https://reddata.atlassian.net/browse/GROK-20419) · reviewer: Dmytro Kovalyov (package author)

[GROK-20419]: https://reddata.atlassian.net/browse/GROK-20419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/Tutorials/CHANGELOG.md
+++ b/packages/Tutorials/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v.next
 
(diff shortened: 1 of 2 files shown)


LaykDimon:
already fixed

PR #3934, opened 2026-07-22
GROK-20452: CVMTests: benchmark cold start via explicit run(await)

alex-aprm:
stop→call auto-start only exists for on-demand containers; the auto queue container is a regular one — the cold-start test now does stop → run(await) → first call and reports container-start and first-call (worker init + loadPackage) separately. Verified on sandbox (5/5 benchmark, 12/12 functional). Ticket: GROK-20452.

Generated with [Claude Code](https://claude.com/claude-code)


alex-aprm:
Superseded by #3935 (clean redo without build artifacts)

PR #3941, opened 2026-07-27
GROK-20379: DBTests: ClickHouseADBC suite

Pospelove:
Adds DBTests coverage for the experimental ADBC/Arrow connector (`grok_connect_adbc`), so the ClickHouse suite runs a second time through the ADBC path and its results can be compared against the JDBC baseline.

## What's in here

**New `ClickHouseADBC` test suite** — a full mirror of the existing ClickHouse suite, pointed at the same `db.datagrok.ai` test database:

- `connections/clickhouse-adbc-db-tests.json` — same server/credentials as `ClickHouseDBTests`, only `dataSource` differs (`ClickHouseADBC`).
- `queries/clickhouse-adbc-pattern-test.sql` — the `mock_data` pattern tests (string/int/date/bool patterns, choices, free-text).
- `queries/clickhouse-adbc-output-test.sql` — per-type output tests, with `clickhouse_output_floats.d42` as the expected-value fixture.
- `src/package-api.ts` — regenerated for the new queries.

**Stack wiring**

- `docker/localhost.docker-compose.yaml` — adds a `grok_connect_adbc` service on port 1235 under its own compose profile, and moves the Datlas config to the list-shaped `connectorsSettings` (already supported on the server) so both connector endpoints are registered:
  ```json
  "connectorsSettings": [
    {"grokConnectHost": "grok_connect", "grokConnectPort": 1234},
    {"grokConnectHost": "grok_connect_adbc", "grokConnectPort": 1235}
  ]
  ```
  The server-wide options (`dataframeParsingMode`, `sambaVersion`, …) move up to the top level, which is where they belong now.
- `.github/workflows/packages.yaml` — starts the `grok_connect_adbc` profile alongside the existing ones.
- Bumps `@datagrok-libraries/test` to `^1.4.0`.

## Notes for the reviewer

- The ADBC connector is experimental, so the category is guarded — the suite is skipped when the connector isn't reachable rather than failing the run.
- The connection reuses the same ClickHouse test DB and credentials as the existing `ClickHouseDBTests` connection; nothing new is exposed.
- The ClickHouse demo DB has been down since ~2026-07-12 and the JDBC ClickHouse tests are currently `skip:`-annotated; this suite will only start producing signal once that DB is restored.


Pospelove:
Duplicate — this content was already merged on Jul 23 via #3937 (db159b551a). Verified: the net tree diff against master shows zero difference in the ClickHouseADBC suite files. Closing and deleting the branch.

PR #3959, opened 2026-08-06
[report 2296] GROK-16674: SequenceTranslator: Delete the pattern selected in the Load block's Pattern combobox instead of the live editor name

LaykDimon:
Renaming a pattern in the Oligo Pattern editor without saving and then clicking the trash icon threw an unhandled "Pattern with name X not found" error, because the Load block's delete button identified its target by the live editor name rather than by the saved pattern selected in that block.

The button and its confirmation dialog now use the pattern selected in the Load block's "Pattern" combobox, so deletion always targets an actually stored pattern.

---
**Diff:** +14/-6 · 3 files

**Verified:** reproduction recipe replayed on the patched build — the original error no longer fires

**Full analysis:** [GROK-16674](https://reddata.atlassian.net/browse/GROK-16674)

[GROK-16674]: https://reddata.atlassian.net/browse/GROK-16674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/SequenceTranslator/CHANGELOG.md
+++ b/packages/SequenceTranslator/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v.next
 
+* 2296: Delete the pattern selected in the Load block's Pattern combobox instead of the live editor name
 * Moved the SequenceTranslator Playwright E2E suite into the package (playwright/); helpers from @datagrok-libraries/test/src/playwright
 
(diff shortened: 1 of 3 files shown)


LaykDimon:
actually looks correct to me, small precise changes as well

PR #3962, opened 2026-08-10
Chem: Add ExcludedSketchers package property to hide sketchers from UI menu

gyptian4life:
## Summary

Adds an `ExcludedSketchers` package property that allows administrators to hide specific sketcher options from the Sketcher's `☰` selection menu without uninstalling them. Excluded sketchers remain available programmatically.

### package.json
- Adds `ExcludedSketchers` string property (comma-separated list, nullable, default empty).

### package.ts
- On init, reads `ExcludedSketchers`, splits/trims to build `DG.chem.excludedSketchers`.
- If the user's stored sketcher is in the excluded list, automatically falls back to the first available non-excluded sketcher.

### chem.ts
- Exports `excludedSketchers: string[]` on the `chem` namespace.
- Filters excluded sketchers out of the `☰` menu items.

### README.md
- Documents the new property: how to set it, valid names, and fallback behaviour.

### sketcher-tests.ts
- Adds tests covering exclusion filtering and fallback sketcher selection.

## How to configure

1. Go to **Manage > Packages > Chem > Context Pane > Settings**.
2. Set **ExcludedSketchers** to a comma-separated list of names (e.g. `OpenChemLib, ChemDraw`).

Valid names: `OpenChemLib`, `Ketcher`, `Marvin`, `ChemDraw` (case-sensitive).

### DIFF
--- a/packages/Chem/src/package.ts
+++ b/packages/Chem/src/package.ts
@@ -200,23 +200,15 @@ async function initChemInt(): Promise<void> {
 
   const sketcherFunctions = DG.Func.find({meta: {role: DG.FUNC_TYPES.MOLECULE_SKETCHER}});
   const sketcherFunc = sketcherFunctions.find((e) => e.name === storedSketcherType || e.friendlyName === storedSketcherType);
-  if (sketcherFunc)
+  if (sketcherFunc && !DG.chem.excludedSketchers.includes(sketcherFunc.friendlyName)) {
     DG.chem.currentSketcherType = sketcherFunc.friendlyName;
-  else {
-    if (!!storedSketcherType) {
+  } else {
+    if (!sketcherFunc && !!storedSketcherType) {
       grok.shell.warning(
         `Package with ${storedSketcherType} function is not installed.Switching to ${DG.DEFAULT_SKETCHER}.`);
     }
-
-    DG.chem.currentSketcherType = DG.DEFAULT_SKETCHER;
-  }
-
-  if (DG.chem.excludedSketchers.includes(DG.chem.currentSketcherType)) {
-    const fallback = sketcherFunctions.find(
-      (f) => !DG.chem.excludedSketchers.includes(f.friendlyName)
-    );
-    if (fallback)
-      DG.chem.currentSketcherType = fallback.friendlyName;
+    const fallback = sketcherFunctions.find((f) => !DG.chem.excludedSketchers.includes(f.friendlyName));
+    DG.chem.currentSketcherType = fallback?.friendlyName ?? DG.DEFAULT_SKETCHER;
   }
 }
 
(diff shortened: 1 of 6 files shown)


packages/Chem/src/tests/sketcher-tests.ts
```diff
@@ -210,3 +210,51 @@ const exampleSmiles = 'CC(C(=O)OCCCc1cccnc1)c2cccc(c2)C(=O)c3ccccc3';
...
+  });
+
+  test('excluded sketcher not in menu list', async () => {
+    const funcs = DG.Func.find({meta: {role: DG.FUNC_TYPES.MOLECULE_SKETCHER}});
+    if (funcs.length < 2) return;
+    const toExclude = funcs[funcs.length - 1].friendlyName;
+    DG.chem.excludedSketchers = [toExclude];
+    const visibleFuncs = funcs.filter((f) => !DG.chem.excludedSketchers.includes(f.friendlyName));
+    expect(visibleFuncs.length, funcs.length - 1);
+    expect(visibleFuncs.every((f) => f.friendlyName !== toExclude), true);
```
drizhina: what does this test check exactly? it does not interact with menu or the logic. it checks if the standard JS filter method works.

packages/Chem/src/package.ts
```diff
@@ -203,6 +210,14 @@ async function initChemInt(): Promise<void> {
 
     DG.chem.currentSketcherType = DG.DEFAULT_SKETCHER;
   }
+
+  if (DG.chem.excludedSketchers.includes(DG.chem.currentSketcherType)) {
```
drizhina: why not combine this check with check above. less lines and clearer logic

packages/Chem/src/tests/sketcher-tests.ts
gyptian4life: Added a new commit to address your comment

packages/Chem/src/package.ts
gyptian4life: Added a new commit to address your comment

packages/Chem/src/tests/sketcher-tests.ts
```diff
@@ -210,3 +210,64 @@ const exampleSmiles = 'CC(C(=O)OCCCc1cccnc1)c2cccc(c2)C(=O)c3ccccc3';
...
+    originalExcluded = [...DG.chem.excludedSketchers];
+  });
+
+  after(async () => {
+    DG.chem.excludedSketchers = originalExcluded;
+    grok.shell.closeAll();
+  });
+
+  test('excluded sketcher not in menu list', async () => {
+    const funcs = DG.Func.find({meta: {role: DG.FUNC_TYPES.MOLECULE_SKETCHER}});
```
drizhina, line 228: I would just remove this test. it does not test anything meaningfull

gyptian4life: @drizhina  what about instead of this one liner I test the DOM: click the icon, wait for the menu popup, assert the excluded name is absent from the rendered items. This would catch an issue if the filter was removed from chem.ts.

Or do you think this is just bunk and should be removed completely?

PR #3974, opened 2026-08-15
[ticket] GROK-20719: Bio: Correct four mismatched Bio menu/dialog strings at their declaration sites

LaykDimon:
> ⚠️ **Code review unresolved.** The automated code-review gate flagged quality concerns the fix phase could not resolve within its revision budget. Review the fix carefully before merging:
>
> - `packages/UsageAnalysis/files/TestTrack/Bio/search-spec.ts:83` — This diff renames two automation hooks — `div-Bio---Search---Subsequence-Search-...` -> `div-Bio---Search---Subsequence-Search...` (from the `top-menu` label change at package.ts:1490) and `dialog-matchWithMonomerLibrary` -> `dialog-Match-with-Monomer-Library` (from the new `name:` at package.ts:180). I verified the derivation against `annotate()` in core/client/d4/lib/src/common/html_utils.dart:458-462 (it maps space and `|` to `-` and leaves `.` alone), so both new values are right, and the two Bio playwright tests were correctly updated. But three executable specs in the same repo still hardcode the old values and are the complete remaining consumer set (grep across /workspace and /workspace/public finds no others): `search-spec.ts:83` does `(document.querySelector('[name="div-Bio---Search---Subsequence-Search-..."]') as HTMLElement).click()` — unguarded, so it throws a TypeError on null; `bio-manage-libraries-crud-spec.ts:277,279` waits 30s on `[name="dialog-matchWithMonomerLibrary"]` and then reads a null dialog; `helm-bio-menu-integration-spec.ts:44` carries the old leaf inside its menu sweep. The fix author disclosed this and argued a pipeline scope gate forbids editing UsageAnalysis — that is a harness constraint, not a reason the breakage is acceptable in the merged change.

Four Bio menu strings contradicted the entries they label: a missing friendly name left a dialog headed with the raw identifier, the Subsequence Search dialog was titled 'Substructure Search', its label had a stray space, and the numbering tooltip promised four schemes while two ship.

Each is corrected at its declaration; Bio's Playwright specs follow the two renamed automation names.

Three UsageAnalysis TestTrack specs hardcode the old names and need the same swaps landed with this change.

---
**Diff:** +16/-14 · 8 files

**Verified:** reproduction recipe replayed on the patched build — the original error no longer fires

**Full analysis:** [GROK-20719](https://reddata.atlassian.net/browse/GROK-20719)

[GROK-20719]: https://reddata.atlassian.net/browse/GROK-20719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/Bio/CHANGELOG.md
+++ b/packages/Bio/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## v.next
 
+* GROK-20719: Correct four mismatched Bio menu/dialog strings at their declaration sites
 * Flow: Added table-aware twins of the `Bio | Transform` entries, which mutate the input frame and return void (or hand the same frame back) — unusable on a pipeline canvas, where the node shows a bare pass-through arrow instead of a named result: Extract Sequence Region, To Atomic Level Column, Molecules to HELM Column, and Split to Monomers Columns (which returns the per-position monomer columns as a table of their own). All share the original implementation
 * Flow: Added Convert Sequence Notation `(table, sequence, targetNotation{fasta,separator,helm}, separator?) -> column`, wrapping `ISeqHandler.convert()`. Notation conversion previously existed only behind a dialog
 * Flow: Added Tag as Macromolecule `(table, sequence) -> column`, which runs sequence detection on a plain string column. Nothing could tag a column built inside a flow, so every semType-dependent bio function silently no-opped on derived data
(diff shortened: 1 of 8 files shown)


LaykDimon:
not a bad PR, but still some flaws - tuned onto it and restarting

PR #3999, opened 2026-08-26
DBTests: accept the lossless bigint to int narrowing in expectTable

alex-aprm:
127 provider tests across Snowflake, Postgresql, MariaDb, MySql, Neo4j and MSSQL have failed with `Column id type expected bigint got int` since GROK-20752 landed.

That behaviour is deliberate: the connector ships a `bigint` column as `int` while every value fits int32 and upgrades it back on the first chunk that overflows (`int8AsInt32`, streaming queries only). `core/docs/features/connector-streaming-optimization/RESULTS_P1.md` WO-1.5 records `row_id` arriving as `int` with an identical fingerprint. The expectations were never updated alongside it.

`expectTable` now accepts `int` where `bigint` is declared. It still compares every value — only the type equality is relaxed, and only in that direction — so a narrowing that actually lost data fails exactly as before.

**Verified in CI** on a full DBTests stand: 316 cases, **127 bigint failures → 0**, no new failures. The 6 that remain are unrelated: 5 docker-container tests that time out on the throwaway stand (they pass on the nightly) and `PostgresqlDates`, which is a separate producer-side change in how Postgres interval text is rendered — I confirmed the `string:tokens` codec round-trips that text byte-identically, so it is not a decoding bug and I have not touched its expectation.


alex-aprm:
Superseded by a clean single-commit branch — this one carried three commits already squash-merged into master, which is what made it conflict.

PR #4065, opened 2026-09-09
grok s migrate: idempotent relation placement, and stop a large push hanging

LaykDimon:
Hardening of `grok s pull/push/migrate` ahead of a customer 1.27 → 1.27 whole-stand move.

## What was wrong

**Placement was not idempotent.** Containment is exclusive on the server (a non-link relation deletes that entity's other non-link rows), so a project the plan called `identical` was skipped for relation writing while the projects that *did* write took its entities away, and a stale `isLink` inherited from the target was deleted by the real owner and never re-examined. Pushing the same bundle twice cut a stand from **273 placed relation rows to 90**, and bled a personal root ~90 rows every push.

**A large push hung silently and indefinitely.** Responses the client never read left their connections checked out; enough of them exhausted undici's pool, after which requests queued *before they started* — where no deadline could reach them. A 15,778-entity push stopped issuing requests entirely, with idle CPU and no error.

## Measured

Three migrations into one 1.27.9 stand, each pushed into what the previous left:

| run | entities | failures | pre-existing relations lost |
|---|---|---|---|
| dev 1.28.0 → 1.27.9 | 5,449 | 46 (0.8%) | 6 of 1,889 |
| release 1.27.9 → 1.27.9 | 1,167 | 0 | 4 of 7,255 |
| public → 1.27.9 | 15,778 | — | 1 of 8,396 |

Three consecutive pushes of the same bundle now leave identical state, and a run repairs what an interrupted one stripped. Every remaining relation gap traces to a prerequisite the tool reports *before* writing — packages not installed, or users absent on the target.

## Also

- A reference the source itself can no longer resolve is recorded at pull time and reported as `skip(dead_on_source)` instead of failing the run (50 such failures → 0 on a real bundle).
- Outside references are resolved against the entity **as the bundle stores it**; scanning the live JSON both hid dead references and left package entities pointing at ids the target has never seen (18 → 92 externals resolved on one bundle).
- An id mapping learned from one target is re-verified before being reused against another.
- Selection, expansion, reference resolution and grant reading overlap their requests; entities the walk would discard are recognised from the listing rather than fetched first, which removed 24 request-deadline timeouts on a dev pull.
- `--type` alias list, `count --filter` caveat, conflict-policy behaviour and bundle statefulness documented in `GROK_S.md`.

## Known, not addressed here

`POST /projects?saveRelations=true` costs ~48s on a stand carrying three stacked migrations, so placement of thousands of projects is slow. It is server-side (the recursive namespace CTE) and needs timing against a clean target before a migration window is planned.

Tests: 276 passing.


LaykDimon:
Follow-up found after opening this PR — worth knowing before anyone plans a migration window.

**The server does not persist `isLink = true`.** Probed directly on a 1.27.9 stand: write a relation set with one row flipped to `isLink: true`, read it back, and the flag is `undefined`.

Consequence for `pushRelations` here: the `corrected` counter compares the target's stored flag against what `ownerOf` says it should be, so for those rows it is non-zero on *every* push — and each one triggers a `POST /projects?saveRelations=true`, measured at **~54s** on a stand holding 25k entities / 19.7k relations.

It is not a pure no-op: adding that correction moved a personal root from 902 to 1,019 relations, because re-writing the set also re-asserts the project's containment and reclaims entities another project had taken. State is stable across three consecutive pushes. So it is a trade-off — roughly 117 recovered relations against ~54s per affected project — not a defect to simply delete.

What it needs before a large migration:
- time `saveRelations` against a *clean* 1.27.9 target; this stand carries three stacked migrations and the namespace recomputation is a recursive CTE over `project_relations_all`, so the 54s is probably inflated here
- decide whether `corrected` should gate on something that can actually converge, given the flag never sticks

Not changing it in this PR: the placement logic is what the idempotency work above stabilised, and this wants measurement on a representative stand rather than a late change.

LaykDimon:
Correcting my earlier comment — I extrapolated from the tail and got it wrong.

`POST /projects?saveRelations=true` is **~0.1 s per relation in the project**, near-linear, not a flat ~54 s:

| relations | write |
|---|---|
| 1 | 0.2 s |
| 10 | 1.1 s |
| 571 | 56.4 s |
| 1017 | 93.9 s |

So a **full placement pass is ~32 minutes** for a stand holding 19.5k relations — fine for a one-time migration. What matters is the distribution, not the project count: 1,274 of 1,293 projects hold ≤10 relations and cost under a second each.

**One project holds 14,538 relations and costs ~24 minutes on its own** — and it is the *pushing account's own root* (`Admin`, `is_root`), which the server grows through `addToUserProject` for every entity a push creates. Migrating as an admin account concentrates most of the placement cost into that single write, and it grows with everything already migrated. Worth knowing when choosing the account a migration runs as.

This makes the `isLink` finding above sharper rather than milder: since the flag never persists, the `corrected` path re-writes the same projects every push — and when one of them is that 14.5k-relation root, it is ~24 minutes of no-op per run. Still a trade-off (the re-write reclaims entities other projects took, worth ~117 relations on one root), but the cost side is now measured.

Nothing here changes the code in this PR.

LaykDimon:
Sharpening the `isLink` note above with two more measurements, and a concrete fix for whoever picks this up.

**`is_link` is server-derived, not client-settable.** On a 1.27.9 stand, of 135 link rows, **135** are for an entity that some other project holds as non-link, and **zero** exist without a container. The server maintains "exactly one container, everyone else links" itself. Writing `isLink: true` in a `saveRelations` payload reads back as `false`.

So `pushRelations`' `corrected` counter compares the target's *derived* flag against what `ownerOf()` computes locally, and writes the difference — which the server then recomputes. Wherever the two disagree it re-fires on every push.

**Why it still matters: the write is not free.** `POST /projects?saveRelations=true` scales with what the target holds, and a migration fills the target as it runs:

| target relations_all | large-project write |
|---|---|
| ~1,400 | 1.5 s |
| ~14,400 | 24 s |
| ~23,500 | 54 s |

A 15.7k-entity push measured here: 12,661 entities saved in ~5 min, then placement at 24 s per write. Starting from a clean stand does not help — by the time placement runs, the stand contains everything just migrated.

**Suggested fix (not made here):** the two directions are not symmetric. Claiming is achievable — writing a containment row wins, because a non-link relation deletes the entity's other non-link rows. Releasing is not — the server derives it and ignores what we send. So `corrected` should only count the *claim* direction: `ownerOf(entity) === this project` **and** the target currently holds it as a link. Drop the release direction, which can never converge.

That should remove most needless writes while keeping the containment reclaim the current behaviour buys (worth ~117 relations on one personal root). It wants an A/B on coverage *and* wall time before landing — which is why I am not making it in this PR.

